PR #33, opened 2024-05-23
fix: track transaction number in changes

montekki:
# What ❔

Logs should track the numbers of transactions that they happen in.

## Checklist

### DIFF
--- a/src/instruction_handlers/precompiles.rs
+++ b/src/instruction_handlers/precompiles.rs
@@ -50,7 +50,8 @@ fn precompile_call(
         let query = LogQuery {
             timestamp: Timestamp(0),
             key: abi.to_u256(),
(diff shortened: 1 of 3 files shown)


src/instruction_handlers/precompiles.rs
```diff
@@ -50,7 +50,7 @@ fn precompile_call(
         let query = LogQuery {
             timestamp: Timestamp(0),
             key: abi.to_u256(),
-            tx_number_in_block: Default::default(),
```
joonazan, line 53: Please don't do this, add a comment instead that says that only the first fields are read by the precompile

joonazan (changes requested):

joonazan (approved):

PR #37, opened 2024-05-25
test: Fuzz execution of a single instruction with afl-fuzz

joonazan:
Replaces stack, program and heap with mocks that just contain one value and crash if multiple addresses are accessed. Fuzzes the execution of a single instruction.

The default settings of libFuzzer don't work for me and it doesn't tell what is going wrong, so I wrote this for afl-fuzz instead.

Afl explores lots of instructions in a matter of seconds, though the crashes it finds are not real ones yet because the test is still missing a state validity check.

The validity check should check that fat pointers are well-formed etc. The test should not be run if the initial state is invalid. The test should panic if the state is no longer valid after running an instruction.


src/single_instruction_test/callframe.rs
```diff
@@ -0,0 +1,75 @@
...
+            stipend: u.arbitrary()?,
+            near_calls: vec![],
+            program: u.arbitrary()?,
+            heap: u.arbitrary()?,
+            aux_heap: u.arbitrary()?,
+            heap_size: u.arbitrary()?,
+            aux_heap_size: u.arbitrary()?,
+            calldata_heap: u.arbitrary()?,
+            heaps_i_am_keeping_alive: vec![], // TODO
+            world_before_this_frame: WorldDiff::default().snapshot(), // TODO
```
montekki, line 33: `TODO`s should either be resolved here or reference a tracking issue.

src/single_instruction_test/print_mock_info.rs
```diff
@@ -0,0 +1,41 @@
+use crate::{callframe::Callframe, State, VirtualMachine};
+
+impl VirtualMachine {
+    pub fn print_mock_info(&self) {
+        // TODO world_diff
```
montekki, line 5: And here the same about `TODO`.

src/single_instruction_test/stack.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    pub fn is_valid(&self) -> bool {
+        is_valid_tagged_value(self.read.value_read)
+            && (self.slot_written.is_none()
+                || is_valid_tagged_value((self.value_written, self.pointer_tag_written)))
+    }
+}
+
+#[derive(Default, Debug)]
+pub struct StackPool {}
```
montekki, line 74:
```suggestion
pub struct StackPool;
```

src/single_instruction_test/callframe.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    pub fn dummy() -> Self {
+        Self {
+            address: H160::zero(),
+            code_address: H160::zero(),
+            caller: H160::zero(),
+            exception_handler: 0,
+            context_u128: 0,
+            is_static: false,
+            stack: StackPool {}.get(),
```
montekki, line 60:
```suggestion
            stack: StackPool.get(),
```

src/single_instruction_test/vm.rs
```diff
@@ -0,0 +1,70 @@
...
+                } else {
+                    vec![]
+                },
+                heaps: u.arbitrary()?,
+                transaction_number: u.arbitrary()?,
+                context_u128: u.arbitrary()?,
+            },
+            settings: u.arbitrary()?,
+            world_diff: Default::default(),
+            stack_pool: StackPool {},
```
montekki, line 61:
```suggestion
            stack_pool: StackPool,
```

src/addressing_modes.rs
```diff
@@ -1,12 +1,16 @@
...
 
 pub(crate) trait Source {
-    fn get(args: &Arguments, state: &mut impl Addressable) -> U256;
-    fn is_fat_pointer(args: &Arguments, state: &mut impl Addressable) -> bool;
+    fn get(args: &Arguments, state: &mut impl Addressable) -> U256 {
+        Self::get_with_pointer_flag(args, state).0
+    }
+    fn get_with_pointer_flag(args: &Arguments, state: &mut impl Addressable) -> (U256, bool) {
+        (Self::get(args, state), false)
+    }
```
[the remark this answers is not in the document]
joonazan: It is if neither of the methods are implemented. I tried writing this with an extra trait first but that isn't possible unfortunately.

src/single_instruction_test/stack.rs
joonazan: I had to change it to this in order to silence clippy.

src/addressing_modes.rs
```diff
@@ -265,28 +258,26 @@ trait StackAddressing {
 }
 
 impl<T: StackAddressing> Source for T {
-    fn get(args: &Arguments, state: &mut impl Addressable) -> U256 {
+    fn get_with_pointer_flag(args: &Arguments, state: &mut impl Addressable) -> (U256, bool) {
```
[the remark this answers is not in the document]
joonazan: I find 2-tuples with two different types palatable.

montekki (approved):

PR #38, opened 2024-05-27
test: storage diffs returned by world_diff

joonazan:


src/world_diff.rs
```diff
@@ -320,3 +320,75 @@ pub struct StorageChange {
...
 const COLD_WRITE_AFTER_WARM_READ_REFUND: u32 = STORAGE_ACCESS_COLD_READ_COST;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn test_storage_changes(initial_values: Vec<(([u8; 20], [u8; 32]), [u8; 32])>, first_changes: Vec<(([u8; 20], [u8; 32]), [u8; 32])>, second_changes: Vec<(([u8; 20], [u8; 32]), [u8; 32])>) {
```
montekki, line 331: This is severely unreadable. The top-level tuple has to become a new `type` if that is possible and formatting has also to be applied (`cargo fmt` is not good with formatting code inside the macros).

Besides the top-level explanation for what this test is doing as well as comments of each steps and their expected outcomes are necessary.

montekki (changes requested):

src/world_diff.rs
joonazan: Making the triple a struct is a good idea. That also solves the problem that I cannot implement Arbitrary for U256.

montekki (approved):

PR #41, opened 2024-06-04
feat: implement pointer masking and kernel mode

montekki:
# What ❔

## Checklist

### DIFF
--- a/src/addressing_modes.rs
+++ b/src/addressing_modes.rs
@@ -71,6 +71,7 @@ pub struct Arguments {
     immediate2: u16,
     pub predicate: Predicate,
     static_gas_cost: u8,
+    requires_kernel_mode: bool,
 }
 
 pub(crate) const L1_MESSAGE_COST: u32 = 156250;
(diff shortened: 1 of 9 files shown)


src/callframe.rs
```diff
@@ -152,6 +152,16 @@ impl Callframe {
                 .map(|f| f.previous_frame_gas)
                 .sum::<u32>()
     }
+
+    pub fn is_kernel_mode(&self) -> bool {
+        address_is_kernel(&self.address)
+    }
+}
+
+pub fn address_is_kernel(address: &H160) -> bool {
```
joonazan, line 161: You can write this as `address < (1 << 16).into()`.

joonazan: Worth checking with compiler explorer what code that produces. We could also save a is_kernel boolean for each frame.

src/fat_pointer.rs
```diff
@@ -31,4 +32,40 @@ impl FatPointer {
     pub fn into_u256(self) -> U256 {
         U256::zero() + unsafe { std::mem::transmute::<FatPointer, u128>(self) }
     }
+
+    pub fn validate(&self, is_fresh: bool) -> FatPointerValidationException {
```
joonazan, line 36: This shouldn't be needed. We already do this when constructing a fat pointer. (Otherwise the fuzzer would find crashes because bad fat pointers are dangerous.)

src/addressing_modes.rs
```diff
@@ -79,14 +80,15 @@ pub(crate) const SLOAD_COST: u32 = 2008;
 pub(crate) const INVALID_INSTRUCTION_COST: u32 = 4294967295;
 
 impl Arguments {
-    pub const fn new(predicate: Predicate, gas_cost: u32) -> Self {
+    pub const fn new(predicate: Predicate, gas_cost: u32, requires_kernel_mode: bool) -> Self {
```
joonazan, line 83: Please check kernel mode in the relevant instructions instead of storing it in arguments.

joonazan (changes requested):

PR #42, opened 2024-06-04
feat: implement kernel mode

joonazan:
- privileged instructions may only be used by kernel addresses
- outside the kernel, fat pointers and pubdata are masked

Because I wanted to use the fuzzer to test this feature and it is tricky to test, I ended up fixing all other discrepancies wrt. zk_evm as well.


joonazan:
Returning your calldata is allowed in kernel mode. I should see if the fuzzer can catch this mistake in vm2.

joonazan:
Stipend and extra costs require kernel mode and system call

PR #43, opened 2024-06-05
feat: exposes necessary methods on heap newtype

montekki:
# What ❔

## Checklist

### DIFF
--- a/src/heap.rs
+++ b/src/heap.rs
@@ -22,7 +22,7 @@ pub struct Heap(Vec<u8>);
 
 impl Heap {
     pub fn reserve(&mut self, additional: usize) {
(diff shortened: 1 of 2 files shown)


src/heap.rs
```diff
@@ -20,6 +20,30 @@ impl HeapId {
 #[derive(Debug, Clone, PartialEq)]
 pub struct Heap(Vec<u8>);
 
+impl Heap {
+    pub fn resize(&mut self, new_len: usize, value: u8) {
```
joonazan, line 24: Do we actually need resize and len?

I would like to keep the backing storage as an implementation detail if possible.

joonazan (approved):

PR #44, opened 2024-06-14
feat: decommit opcode

montekki:
# What ❔

## Checklist


src/program.rs
```diff
@@ -25,6 +25,10 @@ impl Program {
         self.instructions.get::<usize>(n.into())
     }
 
+    pub fn instructions(&self) -> &[Instruction] {
```
joonazan, line 28: Remove this. I already fixed the tracing in a way that doesn't require this.

src/instruction_handlers/decommit.rs
```diff
@@ -0,0 +1,117 @@
...
+                .code_page()
+                .as_ref()
+                .iter()
+                .enumerate()
+            {
+                let mut bytes = [0; 32];
+                word.to_big_endian(&mut bytes);
+                let h: H256 = H256::from(bytes);
+
+                vm.state.heaps[heap].write_u256((i * 32) as u32, *word);
```
joonazan, line 85: should use HeapInterface::memset/from_slice here.

src/decommit.rs
```diff
@@ -94,8 +109,15 @@ impl WorldDiff {
 }
 
 pub(crate) struct UnpaidDecommit {
-    cost: u32,
-    code_key: U256,
+    pub cost: u32,
+    pub code_key: U256,
+    pub is_initial: bool,
```
joonazan, line 114: The is_initial field seems unnecessary.

joonazan: Clippy agrees

afl-fuzz/src/main.rs
```diff
@@ -8,6 +8,9 @@ fn main() {
             if vm.is_in_valid_state() && vm.instruction_is_not_precompile_call() {
                 // Tests that running one instruction and converting to zk_evm produces the same result as
                 // first converting to zk_evm and then running one instruction.
+                vm.state
```
joonazan, line 11: Please do this in the code that generates `vm.state`.

joonazan: (Otherwise you'd need to copy it into show_crash)

src/single_instruction_test/vm.rs
```diff
@@ -40,19 +40,17 @@ impl VirtualMachine {
...
-        // TODO PLA-934 implement Decommit
-        if self.current_opcode() == 1093 {
-            return false;
-        }
-
-        // TODO PLA-972 implement StaticMemoryRead/Write
-        if (1096..=1103).contains(&self.current_opcode()) {
-            return false;
-        }
+        self.current_opcode() == 1093
```
joonazan, line 43: Don't commit this!

src/decommit.rs
```diff
@@ -91,11 +108,21 @@ impl WorldDiff {
         self.decommitted_hashes.insert(decommit.code_key, ());
         Some(world.decommit(decommit.code_key))
     }
+
+    pub(crate) fn unpaid_decommit(
```
joonazan, line 112: No reason to mimic the two function split. Just make one function that does the whole decommit for decommit opcode and leave `UnpaidDecommit` alone.

src/instruction_handlers/decommit.rs
```diff
@@ -0,0 +1,107 @@
...
+        instruction,
+        world,
+        |vm, args, world, continue_normally| {
+            let code_hash = Register1::get(args, &mut vm.state);
+            let extra_cost = Register2::get(args, &mut vm.state).low_u32();
+
+            let mut decommit_preimage_format_is_invalid = false;
+            let mut buffer = [0u8; 32];
+            code_hash.to_big_endian(&mut buffer);
+            let mut _decommit_preimage_normalized = VersionedHashNormalizedPreimage::default();
```
joonazan, line 32: Please get rid of all unused variables or remove the underscores.

joonazan (changes requested):

src/decommit.rs
```diff
@@ -93,9 +104,10 @@ impl WorldDiff {
     }
 }
 
+#[derive(Debug)]
 pub(crate) struct UnpaidDecommit {
-    cost: u32,
-    code_key: U256,
+    pub cost: u32,
```
joonazan, line 109: Remove these changes, too.

src/single_instruction_test/heap.rs
```diff
@@ -47,6 +53,7 @@ impl<'a> Arbitrary<'a> for Heap {
 
 #[derive(Debug, Clone, Arbitrary)]
```
joonazan, line 54: You can no longer derive Arbitrary because `heap_id` should not be arbitrary. Make a constructor that takes heap_id instead.

src/single_instruction_test/vm.rs
```diff
@@ -90,7 +87,7 @@ impl<'a> Arbitrary<'a> for VirtualMachine {
                 // Exiting the final frame is different in vm2 on purpose,
                 // so always generate two frames to avoid that.
                 previous_frames: vec![(0, Callframe::dummy())],
-                heaps: u.arbitrary()?,
+                heaps,
```
joonazan, line 90: Use the constructor mentioned in previous comment here. That will make the handcrafted fuzzer inputs work again. (They break if the random generation changes even a little bit, which sucks but I think we can get away with it)

src/single_instruction_test/heap.rs
```diff
@@ -34,6 +34,12 @@ impl HeapInterface for Heap {
         // This is wrong, but this method is only used to get the final return value.
         vec![]
     }
+
+    fn memset(&mut self, src: &[U256]) {
+        for (i, word) in src.iter().enumerate() {
+            self.write_u256((i * 32) as u32, *word);
```
joonazan, line 40: This would break if the mock decommit returned more than one u256 of code page. But maybe it is ok like this.

src/instruction_handlers/decommit.rs
```diff
@@ -0,0 +1,83 @@
...
+        |vm, args, world, continue_normally| {
+            let code_hash = Register1::get(args, &mut vm.state);
+            let extra_cost = Register2::get(args, &mut vm.state).low_u32();
+
+            let mut buffer = [0u8; 32];
+            code_hash.to_big_endian(&mut buffer);
+
+            let preimage_len_in_bytes =
+                zkevm_opcode_defs::system_params::NEW_KERNEL_FRAME_MEMORY_STIPEND;
+            let mut _decommit_header = VersionedHashHeader::default();
```
joonazan, line 33: This variable still has an underscore.

joonazan (changes requested):

src/instruction_handlers/decommit.rs
```diff
@@ -0,0 +1,80 @@
...
+            let code_hash = Register1::get(args, &mut vm.state);
+            let extra_cost = Register2::get(args, &mut vm.state).low_u32();
+
+            let mut buffer = [0u8; 32];
+            code_hash.to_big_endian(&mut buffer);
+
+            let preimage_len_in_bytes =
+                zkevm_opcode_defs::system_params::NEW_KERNEL_FRAME_MEMORY_STIPEND;
+
+            if vm.state.use_gas(extra_cost).is_err() {
```
joonazan, line 32: This if can be merged with the next one.

src/instruction_handlers/decommit.rs
```diff
@@ -0,0 +1,80 @@
...
+
+            if !ContractCodeSha256Format::is_valid(&buffer) && !BlobSha256Format::is_valid(&buffer)
+            {
+                Register1::set(args, &mut vm.state, U256::zero());
+                return continue_normally;
+            };
+
+            let Some(program) = vm.world_diff.decommit_opcode(world, code_hash) else {
+                Register1::set(args, &mut vm.state, U256::zero());
+                return continue_normally;
```
joonazan, line 45: Apparently this opcode can never panic, so you should use the normal instruction_boilerplate.

src/instruction_handlers/decommit.rs
```diff
@@ -0,0 +1,80 @@
...
+            };
+
+            let heap = vm.state.heaps.allocate();
```
joonazan, line 48: This heap should be deallocated when no longer in use, especially since we do not recycle these. Adding it to heaps_that_this_frame_is_keeping_alive would do the trick IIRC.

src/decommit.rs
```diff
@@ -77,6 +77,17 @@ impl WorldDiff {
         Some((UnpaidDecommit { cost, code_key }, is_evm))
     }
 
+    pub(crate) fn decommit_opcode(
+        &mut self,
+        world: &mut dyn World,
+        code_hash: U256,
+    ) -> Option<Program> {
```
joonazan, line 84: This should not return a Program, just the code page. You also need to change the World interface because we do not want to decode Programs when we actually only need the code page.

src/single_instruction_test/heap.rs
```diff
@@ -60,10 +67,24 @@ impl Heaps {
     }
 
     pub(crate) fn allocate(&mut self) -> HeapId {
-        HeapId(3)
+        self.heap_id
     }
 
     pub(crate) fn deallocate(&mut self, _: HeapId) {}
+
+    pub fn set_heap_id(&mut self, heap_id: HeapId) {
```
joonazan, line 75: Remove this method.

src/single_instruction_test/vm.rs
```diff
@@ -81,6 +76,8 @@ impl<'a> Arbitrary<'a> for VirtualMachine {
             register_pointer_flags |= (is_pointer as u16) << i;
         }
 
+        let heaps = Heaps::from_id(current_frame.heap, u)?;
```
joonazan, line 79: Inline this variable.

joonazan (changes requested):

src/single_instruction_test/vm.rs
montekki: will not work nicely due to `current_frame` moving in constructor.

joonazan (approved):

PR #46, opened 2024-06-25
feat: Stable tracer interface

joonazan:


eravm-stable-interface/src/lib.rs
```diff
@@ -0,0 +1,136 @@
+use u256::{H160, U256};
+
+pub trait StateInterface {
```
slowli, line 3: Dumb question: Is all this functionality required for tracers we want to implement initially? AFAIU, e.g. the call tracer is "readonly" (doesn't change the VM state). What are the examples of tracers mutating the VM state for the old VM? In general, since we don't need to have stable API right away, I'd suggest to have a minimal interface for the known uses cases (call tracer), and then iteratively add / improve on the interface as new tracers are developed.

eravm-stable-interface/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn flags(&self) -> Flags;
+    fn set_flags(&mut self, flags: Flags);
+
+    fn transaction_number(&self) -> u16;
+    fn set_transaction_number(&mut self, value: u16);
+
+    fn context_u128_register(&self) -> u128;
+    fn set_context_u128_register(&mut self, value: u128);
+
+    /// Returns the new value and the amount of pubdata paid for it.
```
slowli, line 23: Is this comment outdated? Or does it belong to `get_storage()`? In general, I'd suggest defining a `struct` for the returned value, so that the purpose of fields is obvious from their name.

src/instruction.rs
```diff
@@ -19,9 +19,8 @@ impl fmt::Debug for Instruction {
     }
 }
 
-pub(crate) type Handler =
-    fn(&mut VirtualMachine, *const Instruction, &mut dyn World) -> InstructionResult;
-pub(crate) type InstructionResult = Result<*const Instruction, ExecutionEnd>;
+pub(crate) type Handler = fn(&mut VirtualMachine, &mut dyn World) -> InstructionResult;
+pub(crate) type InstructionResult = Option<ExecutionEnd>;
```
slowli, line 23: Bikeshedding: It's weird to see a type named "..Result" be an `Option<_>`. I'd rename it `InstructionOutput` or remove the type alias altogether.

src/tracing.rs
```diff
@@ -0,0 +1,364 @@
...
+    }
+}
+
+struct CallframeWrapper<'a> {
+    frame: &'a mut Callframe,
+    near_call: Option<usize>,
+}
+
+impl CallframeInterface for CallframeWrapper<'_> {
+    fn address(&self) -> u256::H160 {
```
slowli, line 184: Bikeshedding: At this point it probably doesn't make sense to rename `primitive_types` in `Cargo.toml` because `U256` is not the only type imported.

eravm-stable-interface/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+
+pub trait StateInterface {
+    fn read_register(&self, register: u8) -> (U256, bool);
+    fn set_register(&mut self, register: u8, value: U256, is_pointer: bool);
+
+    fn number_of_callframes(&self) -> usize;
+    /// zero is the current frame, one is the frame before that etc.
+    fn callframe(&mut self, n: usize) -> impl CallframeInterface + '_;
+
+    fn read_heap_byte(&self, heap: HeapId, index: u32) -> u8;
```
slowli, line 11: Why is it possible to only read / write single bytes at a time? This doesn't look very efficient / ergonomic. IMO, could make more sense to have methods to copy a range of bytes from / to heap, i.e., something like

```rust
fn copy_from_heap(&self, id: HeapId, offset: u32, buffer: &mut [u8]);
fn copy_to_heap(&mut self, id: HeapId, offset: u32, buffer: &[u8]);
```

eravm-stable-interface/src/lib.rs
joonazan: No, we definitely don't need all this.

However, I went over this design pretty carefully with Igor from the compiler team and this is the bare minimum of what they need.

src/instruction.rs
joonazan: Maybe the name is bad.

What's even worse IMO is the Option. I want to change it to something like `Running | Stopped`.

eravm-stable-interface/src/lib.rs
joonazan: The problem with your suggestion is that it has worse performance because it introduces unnecessary copying/allocation. An iterator-based interface might be more ergonomic, though it is also more complex.

Ergonomics and performance are secondary goals here, though. For instance, any sane user would want to know what instruction is currently being executed but that is not part of the interface because the interface is stable and minimal.

To get the currently executed instruction, the user must get the pc, read the code page and then decode that. That can't be done in this library because zkevm_opcode_defs' extensions aren't done in a backwards compatible way. 😢 Because of that complexity, I think another ergonomics layer is warranted.

joonazan: How would you provide that nice interface?

One possibility is making a better zkevm_opcode_defs. I already have code that rips the instruction encoding.
https://github.com/matter-labs/eravm_instruction_encoding

joonazan:
I wrote tests and found out that the type system construction that I did doesn't actually work. The generic parameter and autoref specialization don't work together. I think there is some alternative formulation that does work but I haven't found it yet.

src/tracing.rs
```diff
@@ -0,0 +1,366 @@
...
+            .iter()
+            .map(|frame| frame.near_calls.len() + 1)
+            .sum::<usize>()
+            + self.state.current_frame.near_calls.len()
+            + 1
+    }
+
+    fn callframe(&mut self, n: usize) -> impl CallframeInterface + '_ {
+        for far_frame in std::iter::once(&mut self.state.current_frame)
+            .chain(self.state.previous_frames.iter_mut())
```
slowli, line 37: Dumb question: what is the iteration order here? AFAIU, `previous_frames` have stack ordering (the most recent frame is last), but the current frame is prepended, not appended.

src/tracing.rs
```diff
@@ -0,0 +1,366 @@
...
+        self.state
+            .previous_frames
+            .iter()
+            .map(|frame| frame.near_calls.len() + 1)
+            .sum::<usize>()
+            + self.state.current_frame.near_calls.len()
+            + 1
+    }
+
+    fn callframe(&mut self, n: usize) -> impl CallframeInterface + '_ {
```
slowli, line 35: Another dumb question: What does `n` mean? At first I thought it means an index of far and near call frames considered together, but this doesn't look to be the case; `n` isn't modified when iterating over far call frames, while it would need to be decreased in the `Ordering::Greater` branch of the `match` below. Conceptually, I think it could make sense to treating far and near call frames differently in the VM state interface, since there's a clear semantic distinction between them. This would allow simplifying accessing methods.

eravm-stable-interface/src/tracer_interface.rs
```diff
@@ -0,0 +1,230 @@
...
+    };
+}
+
+// Multiple tracers can be combined by building a linked list out of tuples.
+impl<A: Tracer, B: Tracer> Tracer for (A, B) {
+    forall_opcodes!(dispatch_to_tracer_tuple);
+}
+
+/// Call the before/after_{opcode} method of the tracer.
+pub trait NotifyTracer {
```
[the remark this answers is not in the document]
joonazan: Then we should expose the X macro for generating these kinds of things. We should do that anyway, though.

src/tracing.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    fn pubdata(&self) -> i32 {
+        self.world_diff.pubdata()
+    }
+
+    fn set_pubdata(&mut self, value: i32) {
+        self.world_diff.pubdata.0 = value;
+    }
+
+    fn run_arbitrary_code(_code: &[u64]) {
```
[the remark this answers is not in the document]
joonazan: Raw opcodes. I'll implement the todo! methods eventually but I want to try this out a bit first, so I think I'll have to write some core-side integration.

src/tracing.rs
```diff
@@ -0,0 +1,366 @@
...
+            .map(|(key, value)| (*key, *value))
+    }
+
+    fn get_storage(&self, _address: u256::H160, _slot: u256::U256) -> Option<(u256::U256, u32)> {
+        todo!() // Do we really want to expose the pubdata?
+    }
+
+    fn get_storage_initial_value(&self, _address: u256::H160, _slot: u256::U256) -> u256::U256 {
+        todo!() // Do we really want to expose the caching?
+    }
```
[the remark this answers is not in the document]
joonazan: As you can see I have second thoughts about get_storage_initial_value especially. I'll have to ask them to explain to me again why they want this.

src/tracing.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    fn set_pubdata(&mut self, value: i32) {
+        self.world_diff.pubdata.0 = value;
+    }
+
+    fn run_arbitrary_code(_code: &[u64]) {
+        todo!()
+    }
+
+    fn static_heap(&self) -> HeapId {
```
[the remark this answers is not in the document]
joonazan: It is the heap manipulated by two opcodes that we don't support yet. https://linear.app/matterlabs/issue/PLA-972/implement-staticmemoryreadwrite

These opcodes write to some extra heap that there is only one of. Nothing seems to require them, though.

src/tracing.rs
joonazan: Looks like I need to add some .rev

src/tracing.rs
joonazan: Yes, this code is buggy and I should write a test for it.

I to treat near and far frames as the same thing because what is stored in near and what in far only is pretty vague. This way the API doesn't break if near calls get the ability to mutate some new field.

eravm-stable-interface/src/state_interface.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    /// The VM will execute an invalid instruction if you jump out of the program.
+    fn set_program_counter(&mut self, value: u16);
+
+    fn exception_handler(&self) -> u16;
+
+    fn is_static(&self) -> bool;
+    fn is_kernel(&self) -> bool {
+        todo!()
+    }
```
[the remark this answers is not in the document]
joonazan: In theory it could be implemented in terms of `address` but that is difficult and not particularly important.

src/tracing.rs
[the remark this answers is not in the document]
joonazan: I'm removing these for now, as the compiler team's need for tracers has become unclear.

joonazan:
Removed run_arbitrary_code because of the bizarre control flow that it would create. Having such a feature is desirable but it is better implemented as a thing that can only be used by first suspending the VM.

src/vm.rs
```diff
@@ -59,48 +66,43 @@ impl VirtualMachine {
             ),
             settings,
             stack_pool,
+            panic: Box::new(Instruction::from_panic(
```
slowli, line 69: Dumb question: Would extending `ExecutionStatus` to have an additional "panicked" variant significantly degrade performance? Or would it interfere with plans of tail call optimizations?

eravm-stable-interface/src/tracer_interface.rs
```diff
@@ -0,0 +1,255 @@
...
+macro_rules! pub_struct {
+    ($x:ident) => {
+        pub struct $x;
+    };
+}
+
+pub mod opcodes {
+    forall_simple_opcodes!(pub_struct);
+    pub struct FarCall<const M: u8>;
+    pub struct Ret<const M: u8>;
```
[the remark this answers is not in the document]
joonazan: I thought that this would require some ugly code on the VM side but it actually doesn't if I also switch the VM to this kind of solution. `fn ret<T: Tracer, const RETURN_TYPE: u8,` has the exact same problem where I'd like to have a const enum but that isn't allowed.

eravm-stable-interface/src/tracer_interface.rs
```diff
@@ -0,0 +1,255 @@
...
+impl<const M: u8> OpcodeType for opcodes::Ret<M> {
+    const VALUE: Opcode = Opcode::Ret(ReturnType::from_u8(M));
+}
+
+pub trait Tracer {
+    fn before_instruction<OP: OpcodeType, S: StateInterface>(&mut self, _state: &mut S) {}
+    fn after_instruction<OP: OpcodeType, S: StateInterface>(&mut self, _state: &mut S) {}
+}
+
+/// For example, here `FarCallCounter` counts the number of far calls.
```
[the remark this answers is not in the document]
joonazan: Oops, that got displaced in the refactorings.

eravm-stable-interface/src/tracer_interface.rs
```diff
@@ -0,0 +1,255 @@
...
+///             Opcode::FarCall(_) => self.0 += 1,
+///             _ => {}
+///         }
+///     }
+/// }
+/// ```
+
+impl Tracer for () {}
+
+// Multiple tracers can be combined by building a linked list out of tuples.
```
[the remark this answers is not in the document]
joonazan: You can use this for an arbitrary number of tracers: `(A, (B, C))`. It isn't pretty, though. Adding macros to construct nested tuples imposes no limit on how many tracers you can have but on the other hand supporting tuples up to size N is more intuitive as long as N is big enough.

src/lib.rs
```diff
@@ -46,10 +48,10 @@ use single_instruction_test::stack;
 #[cfg(feature = "single_instruction_test")]
 pub use zkevm_opcode_defs;
 
-pub trait World {
+pub trait World<T> {
```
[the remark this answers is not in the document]
joonazan: The parameter is necessary because dynamic dispatch breaks otherwise. It may be possible to statically dispatch World now, though.

Also, there are probably other solutions. I added the parameter to get something working without analyzing the problem properly.

eravm-stable-interface/src/tracer_interface.rs
```diff
@@ -0,0 +1,255 @@
...
+    This,
+    Caller,
+    CodeAddress,
+    ErgsLeft,
+    SP,
+    ContextMeta,
+    ContextU128,
+    SetContextU128,
+    IncrementTxNumber,
+    AuxMutating0,
```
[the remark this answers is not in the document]
joonazan: I have no idea. This is the official name of the instruction, even though in practice it reads pubdata when in kernel mode. We could maybe rename it.

src/vm.rs
joonazan: Right now `ExecutionStatus::Running` is performing a tail call and `ExecutionStatus::Stopped` is reaching a base case. I'm not sure how the new case would translate.

However, I think it is possible to implement panic more cleanly by doing something like directly calling it.

eravm-stable-interface/src/tracer_interface.rs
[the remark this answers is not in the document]
joonazan: This is now fixed. Never thought of dependent types as a way to avoid menial labor before 😄

eravm-stable-interface/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "eravm-stable-interface"
+version = "0.1.0"
+edition = "2021"
+license = "MIT OR Apache-2.0"
+authors = ["The Matter Labs Team <hello@matterlabs.dev>"]
```
montekki, line 6: Pull these from `workspace`.
```suggestion
version.workspace = true
edition.workspace = true
license.workspace = true
authors.workspace = true
```

src/lib.rs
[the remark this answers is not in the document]
joonazan: Without this parameter the world would have to have program caches for many different tracer types.

src/instruction_handlers/heap_access.rs
```diff
@@ -142,23 +153,25 @@ pub fn grow_heap<H: HeapFromState>(state: &mut State, new_bound: u32) -> Result<
...
-    instruction: *const Instruction,
-    world: &mut dyn World,
-) -> InstructionResult {
-    instruction_boilerplate_with_panic(vm, instruction, world, |vm, args, _, continue_normally| {
+fn load_pointer<T: Tracer, const INCREMENT: bool>(
+    vm: &mut VirtualMachine<T>,
+    world: &mut dyn World<T>,
+    tracer: &mut T,
+) -> ExecutionStatus {
+    instruction_boilerplate::<opcodes::PointerRead, _>(vm, world, tracer, |vm, args, _| {
```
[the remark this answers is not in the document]
joonazan: You can do this in a separate PR.

montekki (approved):

PR #47, opened 2024-07-22
perf: optimize external snapshots

joonazan:
The bootloader only uses a small part of the stack and only changes small areas of its heap, yet both are copied entirely when snapshotting. This fixes that.

### DIFF
--- a/src/callframe.rs
+++ b/src/callframe.rs
@@ -1,5 +1,9 @@
 use crate::{
-    decommit::is_kernel, heap::HeapId, program::Program, stack::Stack, world_diff::Snapshot,
+    decommit::is_kernel,
+    heap::HeapId,
+    program::Program,
+    stack::{Stack, StackSnapshot},
+    world_diff::Snapshot,
     Instruction,
 };
 use u256::H160;
@@ -155,10 +159,66 @@ impl Callframe {
                 .map(|f| f.previous_frame_gas)
                 .sum::<u32>()
     }
+
+    pub(crate) fn snapshot(&self) -> CallframeSnapshot {
+        CallframeSnapshot {
+            stack: self.stack.snapshot(),
+
+            context_u128: self.context_u128,
+            sp: self.sp,
+            gas: self.gas,
+            near_calls: self.near_calls.clone(),
+            heap_size: self.heap_size,
+            aux_heap_size: self.aux_heap_size,
+            heaps_i_am_keeping_alive: self.heaps_i_am_keeping_alive.clone(),
+            world_before_this_frame: self.world_before_this_frame.clone(),
+        }
+    }
+
+    pub fn rollback(&mut self, snapshot: CallframeSnapshot) {
+        let CallframeSnapshot {
+            stack,
+            context_u128,
+            sp,
+            gas,
+            near_calls,
+            heap_size,
+            aux_heap_size,
+            heaps_i_am_keeping_alive,
+            world_before_this_frame,
+        } = snapshot;
+
+        self.stack.rollback(stack);
+
+        self.context_u128 = context_u128;
+        self.sp = sp;
+        self.gas = gas;
+        self.near_calls = near_calls;
+        self.heap_size = heap_size;
+        self.aux_heap_size = aux_heap_size;
+        self.heaps_i_am_keeping_alive = heaps_i_am_keeping_alive;
+        self.world_before_this_frame = world_before_this_frame;
+    }
 }
 
 pub(crate) struct FrameRemnant {
     pub(crate) program_counter: u16,
     pub(crate) exception_handler: u16,
     pub(crate) snapshot: Snapshot,
+}
+
+/// Only contains the fields that can change (other than via tracer).
+pub struct CallframeSnapshot {
+    stack: StackSnapshot,
+
+    context_u128: u128,
+    sp: u16,
+    gas: u32,
+    near_calls: Vec<NearCallFrame>,
+
+    heap_size: u32,
+    aux_heap_size: u32,
+
+    heaps_i_am_keeping_alive: Vec<HeapId>,
+    world_before_this_frame: Snapshot,
 }
--- a/src/callframe.rs
+++ b/src/callframe.rs
@@ -175,7 +175,7 @@ impl Callframe {
         }
     }
 
-    pub fn rollback(&mut self, snapshot: CallframeSnapshot) {
+    pub(crate) fn rollback(&mut self, snapshot: CallframeSnapshot) {
         let CallframeSnapshot {
             stack,
             context_u128,
@@ -208,7 +208,7 @@ pub(crate) struct FrameRemnant {
 }
 
 /// Only contains the fields that can change (other than via tracer).
-pub struct CallframeSnapshot {
+pub(crate) struct CallframeSnapshot {
     stack: StackSnapshot,
 
     context_u128: u128,
--- a/src/callframe.rs
+++ b/src/callframe.rs
@@ -170,11 +170,16 @@ impl Callframe {
             near_calls: self.near_calls.clone(),
             heap_size: self.heap_size,
             aux_heap_size: self.aux_heap_size,
-            heaps_i_am_keeping_alive: self.heaps_i_am_keeping_alive.clone(),
+            heaps_i_was_keeping_alive: self.heaps_i_am_keeping_alive.len(),
         }
     }
 
-    pub(crate) fn rollback(&mut self, snapshot: CallframeSnapshot) {
+    /// Returns heaps that were created during the period that is rolled back
+    /// and thus can't be referenced anymore and should be deallocated.
+    pub(crate) fn rollback(
+        &mut self,
+        snapshot: CallframeSnapshot,
+    ) -> impl Iterator<Item = HeapId> + '_ {
         let CallframeSnapshot {
             stack,
             context_u128,
@@ -183,7 +188,7 @@ impl Callframe {
             near_calls,
             heap_size,
             aux_heap_size,
-            heaps_i_am_keeping_alive,
+            heaps_i_was_keeping_alive,
         } = snapshot;
 
         self.stack.rollback(stack);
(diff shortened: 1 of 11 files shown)


src/callframe.rs
```diff
@@ -155,10 +159,66 @@ impl Callframe {
...
 }
 
 pub(crate) struct FrameRemnant {
     pub(crate) program_counter: u16,
     pub(crate) exception_handler: u16,
     pub(crate) snapshot: Snapshot,
 }
+
+/// Only contains the fields that can change (other than via tracer).
+pub struct CallframeSnapshot {
```
slowli, line 211: Nit: Can't the visibility of this type be reduced to `pub(crate)`? IMO, it's useful to have minimum necessary visibility for types to have immediate understanding what the logical scope of the type is.

src/heap.rs
```diff
@@ -21,8 +21,15 @@ impl HeapId {
...
-        self.0.reserve_exact(additional);
+    fn write_u256(&mut self, start_address: u32, value: U256) {
+        let end = (start_address + 32) as usize;
+        if end > self.0.len() {
+            self.0.resize(end, 0);
+        }
+
+        let mut bytes = [0; 32];
+        value.to_big_endian(&mut bytes);
+        self.0[start_address as usize..end].copy_from_slice(&bytes);
```
slowli, line 32: Isn't it possible to call `to_big_endian` directly on heap bytes?

```rust
value.to_big_endian(&mut self.0[start_address as usize..end]);
```

src/heap.rs
```diff
@@ -73,44 +70,63 @@ impl Heaps {
...
         id
     }
 
     pub(crate) fn deallocate(&mut self, heap: HeapId) {
-        self.0[heap.0 as usize].0 = vec![];
+        self.heaps[heap.0 as usize].0 = vec![];
+    }
+
+    pub fn write_u256(&mut self, heap: HeapId, start_address: u32, value: U256) {
+        if heap == FIRST_HEAP {
```
slowli, line 98: Just to understand better: Is it always the case that *all* writes to the bootloader heap happen when there's a living snapshot (i.e., need to be journaled)?

src/heap.rs
```diff
@@ -73,44 +70,63 @@ impl Heaps {
...
+        }
+        self.heaps[heap.0 as usize].write_u256(start_address, value);
+    }
+
+    pub(crate) fn snapshot(&self) -> usize {
+        self.bootloader_heap_rollback_info.len()
+    }
+
+    pub(crate) fn rollback(&mut self, snapshot: usize) {
+        for (address, value) in self.bootloader_heap_rollback_info.drain(snapshot..) {
```
slowli, line 110: Doesn't the order of slots need to be reversed in the general case to account for multiple writes to the same address?

src/state.rs
```diff
@@ -106,6 +106,40 @@ impl State {
     pub(crate) fn get_context_u128(&self) -> u128 {
         self.current_frame.context_u128
     }
+
+    pub(crate) fn snapshot(&self) -> StateSnapshot {
+        //assert!(self.heaps[self.current_frame.aux_heap].is_empty());
```
slowli, line 111: Please either remove the commented out code or restore it.

src/stack.rs
```diff
@@ -53,6 +53,43 @@ impl Stack {
...
         self.pointer_flags.clear(slot);
     }
+
+    pub(crate) fn snapshot(&self) -> StackSnapshot {
+        let mut dirty_prefix_end = 0;
+        for i in 0..NUMBER_OF_DIRTY_AREAS {
+            if self.dirty_areas & (1 << i) != 0 {
+                dirty_prefix_end = i + 1;
+            }
+        }
```
slowli, line 63: IIUC, this can be optimized using [`leading_zeros`](https://doc.rust-lang.org/std/primitive.u64.html#method.leading_zeros) which translates to a single assembly instruction on x86.

src/heap.rs
```diff
@@ -57,13 +54,13 @@ impl HeapInterface for Heap {
...
-    fn memset(&mut self, src: &[u8]) {
-        self.0 = src.to_vec();
-    }
 }
 
 #[derive(Debug, Clone)]
-pub struct Heaps(Vec<Heap>);
+pub struct Heaps {
+    heaps: Vec<Heap>,
+    bootloader_heap_rollback_info: Vec<(u32, U256)>,
```
slowli, line 62: Does this `Vec` only ever grows during VM execution (i.e., never emptied)? Maybe, it makes sense to introduce explicit snapshot commit (or implicit one somehow?) so that we do have an ability to drop unnecessary snapshot info. AFAIU, we do have this hook in the glue code (`pop_snapshot_no_rollback`).

More generally, could the snapshot lifecycle be simplified so that (external) snapshots are never embedded into each other? (I *think* we have this property rn with how VM is used in Era, but not 100% sure.) If this is the case (or even if it's not), it could make sense to embed external snapshot management into `VirtualMachine` so that it maintains basic invariants internally, such as changes being committed / rolled back in the LIFO order.

src/heap.rs
joonazan: That's a good point. This is the case only in sequencer I think, though sequencer is of course the top priority for performance.

joonazan: At least history can be cleared when the last snapshot is forgotten. I think that happens when a TX succeeds.

src/heap.rs
joonazan: Correct. The code is missing `.rev()`.

src/stack.rs
joonazan: I believe LLVM will remove the loop anyway but this reminds me that I was thinking of maybe saving the dirty parts instead of the dirty prefix. WDYT about that?

src/heap.rs
joonazan: I remember that there was some case with two external snapshots but I don't remember where it was or if it was important.

joonazan:
@slowli Nice review! Looks like I was way too distracted yesterday to write correct code.

src/stack.rs
slowli: Saving dirty parts looks more logical given that we track them, but I'm not sure about performance impact of this decision. Regarding LLVM, it will probably unroll the loop, but I'm not sure it's smart enough to detect simplification to a  "count leading zeros" instruction.

src/stack.rs
```diff
@@ -53,6 +53,43 @@ impl Stack {
...
+            dirty_areas,
+            slots,
+        } = snapshot;
+
+        self.zero();
+
+        self.pointer_flags = pointer_flags;
+        self.dirty_areas = dirty_areas;
+        for (me, snapshot) in self.slots.iter_mut().zip(slots.iter()) {
+            *me = *snapshot;
```
montekki, line 84: Maybe it would make sense to `Hasher` the slots in order to not-overwrite the same values with same values here?

src/heap.rs
slowli: Yep, there are embedded snapshots in `MainBatchExecutor` code:

- [`execute_tx()`](https://github.com/matter-labs/zksync-era/blob/dea6969d1b67c54a0985278de68a8d50f1084dc1/core/node/state_keeper/src/batch_executor/main_executor.rs#L171) takes a snapshot unconditionally and never rolls its back or pops it. IIUC, it's expected that [`rollback_last_tx`](https://github.com/matter-labs/zksync-era/blob/dea6969d1b67c54a0985278de68a8d50f1084dc1/core/node/state_keeper/src/batch_executor/main_executor.rs#L204C8-L204C24) is called after a halted transaction, but it looks concerning that a snapshot doesn't seem to be popped otherwise; I'll investigate whether it's the case.
- [`execute_tx_in_vm_with_optional_compression`](https://github.com/matter-labs/zksync-era/blob/dea6969d1b67c54a0985278de68a8d50f1084dc1/core/node/state_keeper/src/batch_executor/main_executor.rs#L257) takes an additional snapshot to handle compression failures. It rolls back a snapshot or pops it inside the method.

There are no VM modifications between taking these 2 snapshots, which makes me think that it could be possible to avoid embedding here. The only other use of `VmInstance::make_snapshot()` looks to be in [the `vm_utils` crate](https://github.com/matter-labs/zksync-era/blob/dea6969d1b67c54a0985278de68a8d50f1084dc1/core/lib/vm_utils/src/lib.rs#L61), and it follows the linear pattern.

src/stack.rs
joonazan: I verified with godbolt.org. This is indeed a case where LLVM doesn't realize it can emit `bsr`.

src/heap.rs
joonazan: It's worth doing probably but I'd say making the change is out of scope for this PR.

src/stack.rs
joonazan: I'm not sure what you mean. Performance-wise it wouldn't be acceptable to touch a hashmap on every stack write.

src/state.rs
```diff
@@ -106,6 +106,46 @@ impl State {
...
+            flags,
+            bootloader_frame,
+            bootloader_heap_snapshot,
+            transaction_number,
+            context_u128,
+        } = snapshot;
+
+        for heap in self.current_frame.rollback(bootloader_frame) {
+            self.heaps.deallocate(heap);
+        }
```
slowli, line 137: Just to understand better: Would encapsulating deallocation logic in the heap snapshot not work? E.g., snapshotting the number of heaps and deallocating heaps with larger IDs on the rollback. Could it be the case that the bootloader would keep heaps with smaller IDs alive, or that some of the heaps in the rolled back execution need to be retained anyway?

slowli (dismissed):

src/state.rs
joonazan: All heaps created during the rolled back time are deallocated but most of them get deallocated earlier. Only the ones kept alive by the bootloader need to be cleaned up.

Heap numbering is not guaranteed to be sequential, though it currently is.

montekki (approved):

PR #48, opened 2024-07-23
feat: Track `storage_refunds` and `pubdata_costs` stats

slowli:
# What ❔

- Tracks `storage_refunds` and `pubdata_costs` in `WorldDiff` in order to fill in the corresponding data in the VM glue code.
- Fixes decommit opcode semantics, refunding extra gas cost if the decommitted bytecode is not fresh (i.e., was decommitted previously).

## Why ❔

- `storage_refunds` and `pubdata_costs` data is not read directly on the EN, but is persisted for each batch in the state keeper. By design, this data is expected to filled in by the state keeper, so filling it retrospectively could be a problem.
- Divergency in decommit opcode semantics is easily observable (e.g., by gas costs).

fix: Fix decommit opcode semantics

### DIFF
--- a/src/world_diff.rs
+++ b/src/world_diff.rs
@@ -252,8 +252,9 @@ impl WorldDiff {
         self.l2_to_l1_logs.logs_after(snapshot.l2_to_l1_logs)
     }
 
-    pub fn get_decommitted_hashes(&self) -> &BTreeMap<U256, ()> {
-        self.decommitted_hashes.as_ref()
+    /// Returns hashes of decommitted contract bytecodes in no particular order.
+    pub fn decommitted_hashes(&self) -> impl Iterator<Item = U256> + '_ {
+        self.decommitted_hashes.as_ref().keys().copied()
     }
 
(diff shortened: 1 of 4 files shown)


src/world_diff.rs
```diff
@@ -227,27 +268,23 @@ impl WorldDiff {
...
+        self.storage_changes.rollback(snapshot.storage_changes);
+        self.paid_changes.rollback(snapshot.paid_changes);
+        self.events.rollback(snapshot.events);
+        self.l2_to_l1_logs.rollback(snapshot.l2_to_l1_logs);
         self.transient_storage_changes
-            .rollback(transient_storage_changes);
-        self.pubdata.rollback(pubdata);
+            .rollback(snapshot.transient_storage_changes);
+        self.pubdata.rollback(snapshot.pubdata);
+        if rollback_io_history {
```
joonazan, line 284: Looks like this only happens in the external rollback, so please remove the boolean and move the if's body to the external rollback.

joonazan (changes requested):
I don't see any other issues, though I didn't check the decommit changes but I trust they are an improvement.

PR #50, opened 2024-07-29
chore: Update zkEVM dependencies

slowli:
# What ❔

Uses published versions of zkEVM dependencies.

## Why ❔

To align with dependencies used in the Era repo and to pin to a specific dependency commit.

## Checklist


Cargo.toml
```diff
@@ -8,12 +8,12 @@ license = "MIT OR Apache-2.0"
...
 [dependencies]
-zkevm_opcode_defs = { git = "https://github.com/matter-labs/era-zkevm_opcode_defs.git", branch = "v1.5.0" }
-zk_evm_abstractions = {git = "https://github.com/matter-labs/era-zk_evm_abstractions.git", branch = "v1.5.0" }
+zkevm_opcode_defs = "0.150.0"
+zk_evm_abstractions = "0.150.0"
 u256 = { package = "primitive-types", version = "0.12.1" }
 enum_dispatch = "0.3"
 arbitrary = { version = "1", features = ["derive"], optional = true }
-zk_evm = { git = "https://github.com/matter-labs/era-zk_evm.git", branch = "jms-remove-assert", optional = true }
+zk_evm = { version = "0.150.0", optional = true } # FIXME: update `jms-remove-assert`?
```
slowli, line 16: @joonazan Is removing an assertion necessary for code in the repo to work? (I suppose fuzzing, right?)

joonazan: That assert can be triggered by correct code and has no actual function as far as I can see. It is unlikely to trigger outside of fuzzing but really should be removed.

joonazan: I requested some reviews on it, let's see if it goes in.

joonazan: Looks like we're getting a version of the assertion that is correct today. Don't know when it will be released, though.

slowli: The PR is still not merged... Do you think it would be acceptable to merge this PR as-is given that `zk_evm` is only used for tests and bump its revision later? Unless `zkevm_opcode_defs` and `zk_evm_abstractions` are updated, we won't be able to merge the VM feature branch in the Era codebase.

joonazan: I have some hope that we'll get zk_evm updated. I think we can keep this open until we're blocked by it.

joonazan (approved):

PR #51, opened 2024-07-31
fix: Fix `Heap` equality comparison

slowli:
# What ❔

- Fixes `Heap` equality comparison to allow for additional allocation after a rollback.
- Shortens `Debug` output for large amounts of data (e.g., `Stack`, `Heap`, `Program`) and adds `Debug` impl for a couple of types.

## Why ❔

- Equality comparison is used in `zksync_multivm` unit tests.
- More readable `Debug` implementations simplify debugging.

## Checklist

- [x] Code has been formatted.

### DIFF
--- a/src/heap.rs
+++ b/src/heap.rs
@@ -49,7 +49,7 @@ impl PartialEq for Heap {
         if shorter_bytes.len() > longer_bytes.len() {
             mem::swap(&mut shorter_bytes, &mut longer_bytes);
         }
-        *shorter_bytes == longer_bytes[..shorter_bytes.len()]
+        shorter_bytes[..] == longer_bytes[..shorter_bytes.len()]
             && longer_bytes[shorter_bytes.len()..]
                 .iter()
                 .all(|&byte| byte == 0)
@@ -171,7 +171,8 @@ impl Index<HeapId> for Heaps {
     }
 }
 
-// Since we never remove `Heap`s (even after rollbacks), we allow additional empty heaps at the end of `Heaps`.
+// Since we never remove `Heap` entries (even after rollbacks – although we do deallocate heaps in this case),
+// we allow additional empty heaps at the end of `Heaps`.
 impl PartialEq for Heaps {
     fn eq(&self, other: &Self) -> bool {
         for i in 0..self.heaps.len().max(other.heaps.len()) {
--- a/src/heap.rs
+++ b/src/heap.rs
@@ -1,5 +1,8 @@
-use crate::instruction_handlers::HeapInterface;
-use std::ops::{Index, Range};
+use crate::{hash_for_debugging, instruction_handlers::HeapInterface};
+use std::{
+    fmt, mem,
+    ops::{Index, Range},
+};
 use u256::U256;
 use zkevm_opcode_defs::system_params::NEW_FRAME_MEMORY_STIPEND;
 
@@ -19,6 +22,23 @@ impl HeapId {
 
 #[derive(Clone)]
 pub struct Heap(Vec<u8>);
+
+impl fmt::Debug for Heap {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        const MAX_DEBUGGED_LEN: usize = 256;
+
+        if self.0.len() <= MAX_DEBUGGED_LEN {
+            formatter.debug_tuple("Heap").field(&self.0).finish()
+        } else {
+            formatter
+                .debug_struct("Heap")
+                .field("len", &self.0.len())
+                .field("start", &&self.0[..MAX_DEBUGGED_LEN])
+                .field("hash", &hash_for_debugging(&self.0))
+                .finish_non_exhaustive()
+        }
+    }
+}
 
 // Two heaps are considered equal even if one of them has greater size allocated (provided that all additional bytes
 // are zeroed). This is to account for the case when a bootloader heap is rolled back; heap is never shrunk.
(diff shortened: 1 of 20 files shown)


montekki (dismissed):

src/heap.rs
```diff
@@ -17,9 +20,42 @@ impl HeapId {
...
+        let mut longer_bytes = &other.0;
+        if shorter_bytes.len() > longer_bytes.len() {
+            mem::swap(&mut shorter_bytes, &mut longer_bytes);
+        }
+        *shorter_bytes == longer_bytes[..shorter_bytes.len()]
+            && longer_bytes[shorter_bytes.len()..]
+                .iter()
+                .all(|&byte| byte == 0)
+    }
+}
```
joonazan, line 57: This is ok but I found it unsatisfying. WDYT about this?

```
impl PartialEq for Heap {
    fn eq(&self, other: &Self) -> bool {
        let mut to_compare = self.0.iter().zip(&other.0);
        let compared_len = to_compare.len();

        to_compare.all(|(a, b)| a == b)
            && self.0[compared_len..].iter().all(|&byte| byte == 0)
            && other.0[compared_len..].iter().all(|&byte| byte == 0)
    }
}
```

src/heap.rs
```diff
@@ -135,6 +171,7 @@ impl Index<HeapId> for Heaps {
     }
 }
 
+// Since we never remove `Heap`s (even after rollbacks), we allow additional empty heaps at the end of `Heaps`.
```
joonazan, line 174: Rollbacks do deallocate them, though, and we could store a linked list in the deallocated heaps and use that to reuse them. Maybe state more explicitly that they are not in use even though they still exist.

src/single_instruction_test/program.rs
```diff
@@ -36,6 +36,10 @@ impl<'a> Arbitrary<'a> for Program {
 }
 
 impl Program {
+    pub fn new(_instructions: Vec<Instruction>, _code_page: Vec<U256>) -> Self {
+        unimplemented!("Not used during fuzzing")
```
joonazan, line 40: Then why add the method?

src/stack.rs
```diff
@@ -1,15 +1,55 @@
...
+    alloc::{alloc, alloc_zeroed, Layout},
+    fmt,
+};
 use u256::U256;
 
-#[derive(PartialEq, Debug)]
+/// Helper wrapper for debugging [`Stack`] / [`StackSnapshot`] contents.
+struct StackStart<I>(I);
+
+impl<I: Iterator<Item = (bool, U256)> + Clone> fmt::Debug for StackStart<I> {
```
joonazan, line 11: I'd put all the Debug related code at the bottom of the file, as it is least relevant for how the code works.

src/world_diff.rs
```diff
@@ -369,11 +370,14 @@ mod tests {
             first_changes in arbitrary_storage_changes(),
             second_changes in arbitrary_storage_changes(),
         ) {
-            let mut world_diff = WorldDiff::default();
-            world_diff.storage_initial_values = initial_values
+            let storage_initial_values = initial_values
```
joonazan, line 373: Nice! I'd inline this variable, too.

joonazan (changes requested):
Many good cleanups overall.

src/world_diff.rs
slowli: IMO, this would make the code less readable (I understand this is completely subjective, so if you insist, I can do this). If I understand Rust compilation correctly, an additional local var leads to the same output at least with optimizations enabled.

src/single_instruction_test/program.rs
slowli: Needed for compilation of integration tests in the main library to succeed. There's a way to signal that a test doesn't work with a particular feature (add `#![cfg(not(feature = ".."))]` at the start of the file), but since there are other dummy methods added to make the library itself work, I've decided that this would be a preferred approach. WDYT? I don't think that having integration tests not compiling at all is a reasonable option, at least when we decide to publish the library.

src/heap.rs
slowli: There are specialized implementations for comparing slices in the `core` ([this one](https://github.com/rust-lang/rust/blob/83dcdb3a5dad0ed1e3e1fadc848d3f7727b41aa5/library/core/src/slice/cmp.rs#L75) is probably relevant for this case), so I'm inclined to politely disagree with a suggestion – zipping and comparing is worse in terms of performance. Subjectively, it's less readable as well. Note that [another, more general impl](https://github.com/rust-lang/rust/blob/83dcdb3a5dad0ed1e3e1fadc848d3f7727b41aa5/library/core/src/slice/cmp.rs#L61) doesn't use zipping for performance reasons 🙃

joonazan: I actually compared the resulting assembly. It is certainly different. Mine has a lot more branching but seems to be shorter an use less stack.

Anyway, we shouldn't be discussing performance as this is only used in tests. I just find this kind of comparison always unappealing, so I tried to find an alternative.

slowli: Being only used in tests is not a part of `PartialEq` contract, and I'd argue that with the amount of data in the bootloader heap, performance is somewhat of a concern even for tests 🙃 If you insist, I can apply your suggestion, but I really don't understand what's so unappealing about comparing slices.

joonazan: Comparing slices is good. Swapping the inputs with mem::swap I don't like but Rust doesn't have .min_max(), so not much that can be done about it.

joonazan (approved):

PR #53, opened 2024-08-09
perf: Implement segmented heap

slowli:
# What ❔

Another stab to segmented heap, similar to how it's implemented for old VMs (+ recycling heap pages).

## Why ❔

Leads to performance improvement compared to the current linear heap design, esp. for single-transaction execution.

## Checklist

### DIFF
--- a/src/heap.rs
+++ b/src/heap.rs
@@ -1,8 +1,7 @@
 use crate::instruction_handlers::HeapInterface;
+use std::mem;
 use std::ops::{Index, Range};
-use std::{iter, mem};
 use u256::U256;
-use zkevm_opcode_defs::system_params::NEW_FRAME_MEMORY_STIPEND;
 
 #[derive(Copy, Clone, PartialEq, Debug)]
 pub struct HeapId(u32);
@@ -52,21 +51,6 @@ impl Heap {
                     HeapPage(boxed_slice.try_into().unwrap())
                 })
             })
-            .collect();
-        Self { pages }
-    }
-
-    fn with_capacity(capacity: usize, recycled_pages: &mut Vec<HeapPage>) -> Self {
-        let end_page = capacity.saturating_sub(1) >> 12;
-        let page_count = end_page + 1;
-        let new_len = recycled_pages.len().saturating_sub(page_count);
-        let recycled_pages = recycled_pages.drain(new_len..).map(|mut page| {
-            page.0.fill(0);
-            Some(page)
-        });
-        let pages = recycled_pages
-            .chain(iter::repeat_with(|| Some(HeapPage::default())))
-            .take(page_count)
             .collect();
         Self { pages }
     }
@@ -216,12 +200,7 @@ impl Heaps {
     }
 
     pub(crate) fn allocate(&mut self) -> HeapId {
-        let id = HeapId(self.heaps.len() as u32);
-        self.heaps.push(Heap::with_capacity(
-            NEW_FRAME_MEMORY_STIPEND as usize,
-            &mut self.recycled_pages,
-        ));
-        id
+        self.allocate_inner(&[])
     }
 
     pub(crate) fn allocate_with_content(&mut self, content: &[u8]) -> HeapId {
@@ -421,13 +400,6 @@ mod tests {
     fn heap_read_out_of_bounds() {
         let heap = Heap::default();
         assert_eq!(heap.read_u256(5), 0.into());
-    }
-
-    #[test]
-    fn default_new_heap_does_not_allocate_many_pages() {
-        let heap = Heap::with_capacity(NEW_FRAME_MEMORY_STIPEND as usize, &mut vec![]);
-        assert_eq!(heap.pages.len(), 1);
-        assert_eq!(*heap.pages[0].as_ref().unwrap().0, [0_u8; 4096]);
     }
 
     fn test_creating_heap_from_bytes(recycled_pages: &mut Vec<HeapPage>) {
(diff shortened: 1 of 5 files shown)


slowli:
Benchmarking results are reported on Linear. TL;DR is that this approach looks consistently faster than the current one (from ~20% on single-transaction benches to ~5% on large batches), and is almost always marginally faster than the previous segmented heap implementation.

src/heap.rs
```diff
@@ -93,6 +150,7 @@ impl HeapInterface for Heap {
 #[derive(Debug, Clone)]
 pub struct Heaps {
     heaps: Vec<Heap>,
+    recycled_pages: Vec<HeapPage>,
```
joonazan, line 153: It is possible to eliminate this Vec entirely by reusing the unused memory as a linked list.
```suggestion
    recycled_pages: Option<Box<UnusedHeapPage>>,
```
where UnusedHeapPage is something like
```rust
struct UnusedHeapPage {
    next: Option<Box<UnusedHeapPage>>,
    extra_data: [u8; HEAP_PAGE_SIZE - 8],
}
```

src/heap.rs
```diff
@@ -20,71 +18,130 @@ impl HeapId {
...
-        for (i, j) in (range.start as usize..end).enumerate() {
-            bytes[i] = self.0[j];
+        if len_in_page < range.len() {
+            if let Some(page) = self.page(page_idx + 1) {
+                result[len_in_page..range.len()]
+                    .copy_from_slice(&page.0[..range.len() - len_in_page]);
+            }
         }
-        U256::from_big_endian(&bytes)
+        U256::from_big_endian(&result)
```
joonazan, line 127: We should definitely store the data in little-endian byteorder. In the Vec-based version that would've required rewriting Vec, but here it is relatively trivial. Just reverse the indices of pages.

joonazan (commented):
I added some suggestions, which may not be very impactful but they should at least improve performance with no downsides.

joonazan:
`read_u256_partially` would be better with `start` and `length` like in @slowli's change that never got merged. Maybe we should switch to that even though it requires coordination with core.

src/heap.rs
```diff
@@ -186,20 +273,245 @@ impl PartialEq for Heaps {
     }
 }
 
+#[derive(Default, Clone, Debug)]
```
slowli, line 276: It's technically sound, but I don't like that `Clone` is required here; AFAIU, it's only used for testing purposes when cloning VM `State` to check whether it's correctly rolled back later. Semantically, it makes little sense to clone the page pool. Could be dealt with later.

src/heap.rs
```diff
@@ -20,71 +17,149 @@ impl HeapId {
...
-        for (i, j) in (range.start as usize..end).enumerate() {
-            bytes[i] = self.0[j];
+        let mut result = [0u8; 32];
+        if let Some(page) = self.page(page_idx) {
+            for (res, src) in result[..bytes_in_page]
+                .iter_mut()
+                .zip(&page.0[offset_in_page..])
+            {
+                *res = *src;
+            }
```
slowli, line 138: Dumb question: Here and elsewhere, are bytewise assignment actually faster than `copy_from_slice`? This is surprising to me (intuitively, `copy_from_slice` would lead to a single non-overlapping memcpy).

slowli (commented):
So, to summarize: `U256`s are still stored in the heap in big-endian form, right?

> Maybe we should switch to that even though it requires coordination with core.

Unless I'm missing something, coordination is well-contained in this case and doesn't require multi-step process, so I'm for it.

src/heap.rs
```diff
@@ -186,20 +273,245 @@ impl PartialEq for Heaps {
...
+    fn recycle_page(&mut self, page: HeapPage) {
+        self.stack_top = Some(DormantHeapPage::from_heap_page(page, self.stack_top.take()));
+    }
+}
+
+#[derive(Debug, Clone)]
+struct DormantHeapPage {
+    next: Option<Box<DormantHeapPage>>,
+    _padding: [u8; HEAP_PAGE_SIZE - mem::size_of::<Option<Box<DormantHeapPage>>>()],
+}
```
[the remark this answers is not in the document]
joonazan: It is very dangerous indeed. Alignment is unlikely to be a problem in practice but there must be some cleaner way of doing this.

src/heap.rs
joonazan: This is correct. I don't like it either

src/heap.rs
joonazan: I haven't checked the assembly but I think this may help the compiler to directly write the data backwards instead of first copying it and then reversing. The loop should result in memcpy anyway if appropriate.

src/heap.rs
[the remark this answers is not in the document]
joonazan: I think we should keep the PagePool refactoring but discard the unsafe optimization for now. It isn't worth it.

src/heap.rs
```diff
@@ -20,71 +17,149 @@ impl HeapId {
...
+
     fn read_range_big_endian(&self, range: Range<u32>) -> Vec<u8> {
-        let end = (range.end as usize).min(self.0.len());
-        let mut result = vec![0; range.len()];
-        for (i, j) in (range.start as usize..end).enumerate() {
-            result[i] = self.0[j];
+        let offset = range.start as usize;
+        let length = (range.end - range.start) as usize;
+
+        let (mut page_idx, mut offset_in_page) = (offset >> 12, offset & (HEAP_PAGE_SIZE - 1));
```
slowli, line 152: Nit: Is there a particular reason `address_to_page_offset()` isn't used here?

slowli (commented):
Nice job! Can't approve this PR since I'm the author 🙃 Feel free to approve it yourself and merge.

joonazan (approved):

PR #54, opened 2024-08-09
feat: statistics for circuit tracer

montekki:
Statistics counting code implemented without the new tracer interface. Later has to be moved there. Only problem so far:
unimplemented static memory instructions lead to divergency on `zksync_state_keeper batch_executor::tests::bootloader_out_of_gas_for_any_tx` test because the way this patch substitutes unimplemented instructions with nop opcodes is wrong. This problem is supposed to go away with the new tracer interface hopefully.


montekki:
closing in favor of #55

PR #55, opened 2024-08-14
chore: fvs circuit statistics tracers interface

montekki:
# What ❔

## Checklist

- [ ] Code has been formatted via `cargo fmt` and `cargo clippy`.


src/decommit.rs
```diff
@@ -94,14 +94,25 @@ impl WorldDiff {
...
+            .decommitted_hashes
+            .insert(decommit.code_key, ())
+            .is_some();
+        let decommit = world.decommit(decommit.code_key);
+
+        let decommit_cycles = if !is_old {
+            (decommit.code_page().len() + 1) / 2
+        } else {
+            0
+        };
```
joonazan, line 113: It seems to me that this could be counted based on `decommitted_hashes` as each entry there is counted once.

External rollbacks interact with this, though. It looks like we'd need to count the cycles of the decommits that are undone by a rollback.

montekki: Only if that becomes a hashmap where the value would be the length of the code page?

joonazan: Right, the length is not readily available.

src/instruction_handlers/storage.rs
```diff
@@ -52,6 +61,18 @@ fn sload<T: Tracer>(
...
         let key = Register1::get(args, &mut vm.state);
+
+        if !vm
+            .world_diff
+            .read_storage_slots_ct
+            .contains(&(vm.state.current_frame.address, key))
+            && !vm
+                .world_diff
+                .written_storage_slots_ct
+                .contains(&(vm.state.current_frame.address, key))
```
joonazan, line 72: Same condition twice.

src/world_diff.rs
```diff
@@ -102,6 +108,7 @@ impl WorldDiff {
             .copied()
             .unwrap_or_else(|| world.read_storage(contract, key).unwrap_or_default());
 
+        self.read_storage_slots_ct.add((contract, key));
```
joonazan, line 111: Instead of adding this new field, you can just change the code so that free storage slots are inserted into `read_storage_slots`, too.

src/world_diff.rs
```diff
@@ -129,6 +136,7 @@ impl WorldDiff {
             .entry((contract, key))
             .or_insert_with(|| world.read_storage(contract, key));
 
+        self.written_storage_slots_ct.add((contract, key));
```
joonazan, line 139: Same thing with written storage slots.

src/instruction_handlers/storage.rs
montekki: no it is not the same it needs to check both read and write storage slots

src/instruction_handlers/storage.rs
```diff
@@ -16,6 +17,14 @@ fn sstore<T: Tracer, W: World<T>>(
     instruction_boilerplate::<opcodes::StorageWrite, _, _>(vm, world, tracer, |vm, args, world| {
         let key = Register1::get(args, &mut vm.state);
         let value = Register2::get(args, &mut vm.state);
+        if !vm
+            .world_diff
+            .written_storage_slots
+            .contains(&(vm.state.current_frame.address, key))
```
joonazan, line 23: These checks add significant overhead and are pretty ugly. Would it be possible to instead look at the written storage slots at the end? All checks seem to just trigger on the first time something is added, except the one check that checks both reads and writes.

eravm-stable-interface/src/state_interface.rs
```diff
@@ -82,6 +83,8 @@ pub trait CallframeInterface {
     fn set_aux_heap_bound(&mut self, value: u32);
 
     fn read_code_page(&self, slot: u16) -> U256;
+
+    fn last_precompile_cycles(&self) -> usize;
```
joonazan, line 87: This is probably unused, as the integration uses the specific precompile cycle numbers instead.

joonazan (changes requested):

joonazan (approved):

joonazan (changes requested):
I think the Decommit opcode isn't tested because tests used to pass even though that one incorrectly incremented the storage cycle count.

It may need to record decommit cycles like far call does but hard to say without a test.

src/state.rs
```diff
@@ -193,3 +196,18 @@ pub(crate) struct StateSnapshot {
...
+#[derive(Debug, Clone, Default)]
+pub struct CycleCounts {
+    pub keccak256_cycles: usize,
+    pub ecrecover_cycles: usize,
+    pub sha256_cycles: usize,
+    pub secp256v1_verify_cycles: usize,
+    pub code_decommitter_cycles: usize,
+}
+
+impl PartialEq for CycleCounts {
```
slowli, line 208: I'd argue that this is an extremely weird impl. It would be better to exclude `cycle_counts` field from the `PartialEq` impl for `State` given that after my PR, there's a manual impl already.

src/instruction_handlers/precompiles.rs
```diff
@@ -66,16 +66,20 @@ fn precompile_call<T: Tracer, W>(
         let heaps = &mut vm.state.heaps;
         match address_low {
             KECCAK256_ROUND_FUNCTION_PRECOMPILE_ADDRESS => {
-                keccak256_rounds_function::<_, false>(0, query, heaps);
+                vm.state.cycle_counts.keccak256_cycles +=
+                    keccak256_rounds_function::<_, false>(0, query, heaps).0;
```
slowli, line 70: Dumb question / maybe was discussed already: Would adding cycle tracking into the tracer functionality be inferior? I.e., something like this:

```rust
pub trait Tracer {
    // ...
    fn after_precompile(&mut self, precompile_stats: PrecompileStats) {}
}

pub enum PrecompileStats {
    Keccak256(usize),
    Sha256(usize),
    EcRecover(usize),
    Secp256k1Verify(usize),
    Decommit(usize),
}
```

src/rollback.rs
```diff
@@ -64,15 +64,48 @@ impl<K: Ord, V> AsRef<BTreeMap<K, V>> for RollbackableMap<K, V> {
...
+pub struct RollbackableSet<K: Ord> {
+    map: BTreeMap<K, ()>,
+    old_entries: Vec<K>,
+}
 
 impl<T: Ord + Clone> RollbackableSet<T> {
-    pub fn add(&mut self, key: T) {
-        self.insert(key, ());
+    /// Adds `key` to the set and returns if it was already present.
+    pub fn add(&mut self, key: T) -> bool {
```
slowli, line 75: Bikeshedding: Suggest to reverse the returned value (and maybe rename the method to `insert`) given that `BTreeSet` / `HashSet` return `true` from `insert` if the value is new.

src/rollback.rs
```diff
@@ -64,15 +64,48 @@ impl<K: Ord, V> AsRef<BTreeMap<K, V>> for RollbackableMap<K, V> {
...
+
+impl<K: Ord> Rollback for RollbackableSet<K> {
+    type Snapshot = usize;
+
+    fn snapshot(&self) -> Self::Snapshot {
+        self.old_entries.len()
+    }
+
+    fn rollback(&mut self, snapshot: Self::Snapshot) {
+        for k in self.old_entries.drain(snapshot..).rev() {
```
slowli, line 96: Nit: does't look like `rev`ersing values matters here.

src/world_diff.rs
```diff
@@ -26,11 +27,13 @@ pub struct WorldDiff {
 
     // The fields below are only rolled back when the whole VM is rolled back.
     pub(crate) decommitted_hashes: RollbackableSet<U256>,
-    read_storage_slots: RollbackableSet<(H160, U256)>,
-    written_storage_slots: RollbackableSet<(H160, U256)>,
+    pub read_storage_slots: RollbackableSet<(H160, U256)>,
+    pub written_storage_slots: RollbackableSet<(H160, U256)>,
```
slowli, line 31: Dumb question: Why are these fields public? AFAIU, they are not accessed in the Era codebase, only `storage_application_cycles` is.

src/world_diff.rs
```diff
@@ -26,11 +27,13 @@ pub struct WorldDiff {
...
 
     // This is never rolled back. It is just a cache to avoid asking these from DB every time.
     storage_initial_values: BTreeMap<(H160, U256), Option<U256>>,
+
+    pub storage_application_cycles: u32,
```
slowli, line 36: If we want to retain stats in this repo, I'd much rather maintain separate read and write stats so that they can be easily aggregated by the caller. Also, I *really* don't like exposing this data as public writeable fields. IMO, interface like `take_storage_cycles(&mut self) -> (u32, u32)` resetting the stats to zero would be preferable (same for precompile stats if they are retained inside the VM).

src/world_diff.rs
joonazan: Noticed the same while rebasing. Should be fixed now.

src/instruction_handlers/common.rs
```diff
@@ -3,28 +3,67 @@ use crate::{addressing_modes::Arguments, instruction::ExecutionStatus, VirtualMa
...
-    instruction_boilerplate_ext::<Opcode, T, W>(vm, world, tracer, |vm, args, _, world| {
+    full_boilerplate::<Opcode, T, W>(vm, world, tracer, |vm, args, world, _| {
         business_logic(vm, args, world);
         ExecutionStatus::Running
     })
 }
 
 #[inline(always)]
-pub(crate) fn instruction_boilerplate_ext<Opcode: OpcodeType, T: Tracer, W>(
+pub(crate) fn boilerplate_t<Opcode: OpcodeType, T: Tracer, W>(
```
slowli, line 32: Bikeshedding / dumb question: Would using `boilerplate_wt` instead of `boilerplate_w` / `boilerplate_t` degrade performance? Subjectively, introducing 2 additional functions isn't really justified here; it's easy to use `_` for unused args, and it's not like there's a chance to mix up tracer and world args.

eravm-stable-interface/src/tracer_interface.rs
```diff
@@ -208,6 +208,19 @@ impl<T: opcodes::TypeLevelReturnType> OpcodeType for opcodes::Ret<T> {
 pub trait Tracer {
     fn before_instruction<OP: OpcodeType, S: StateInterface>(&mut self, _state: &mut S) {}
     fn after_instruction<OP: OpcodeType, S: StateInterface>(&mut self, _state: &mut S) {}
+
+    fn on_extra_prover_cycles(&mut self, _stats: CycleStats) {}
```
slowli, line 212: Have you checked whether implementations are inlined as expected (i.e., there's no performance regression compared to the previous inlined stats implementation)?

slowli (commented):
Other than a couple of nits (including a previous one regarding `RollbackableSet` interface), and the obvious question regarding performance, looks good 👍

eravm-stable-interface/src/tracer_interface.rs
joonazan: I'm assuming there isn't, as at least the integration branch iai slightly improved.

src/instruction_handlers/common.rs
joonazan: Sounds reasonable.

joonazan (approved):

PR #56, opened 2024-08-27
fix: Fix decommitment logic on out-of-gas

slowli:
# What ❔

Records a decommitment event even if it would fail with an out-of-gas error.

## Why ❔

Reproduces old VM behavior.

## Checklist


montekki (approved):

joonazan (approved):
This probably fixes test failures that I couldn't figure out yesterday!

PR #57, opened 2024-09-02
fix: Fix decommitment cost divergence

slowli:
# What ❔

Fixes decommitment cost divergence introduced after fixing the previous VM divergence.

## Why ❔

As of now, bytecode decommitment cost is now not charged even if decommitment failed; this is incorrect. The easiest fix implemented in this PR is to track decommitment status (successful / failed) for all bytecodes.

## Checklist

### DIFF
--- a/eravm-stable-interface/src/tracer_interface.rs
+++ b/eravm-stable-interface/src/tracer_interface.rs
@@ -53,17 +53,18 @@ macro_rules! pub_struct {
 }
 
 pub mod opcodes {
+    use super::{CallingMode, ReturnType};
+    use std::marker::PhantomData;
+
     forall_simple_opcodes!(pub_struct);
-    pub struct FarCall<M: TypeLevelCallingMode>(M);
-    pub struct Ret<T: TypeLevelReturnType>(T);
+    pub struct FarCall<M: TypeLevelCallingMode>(PhantomData<M>);
+    pub struct Ret<T: TypeLevelReturnType>(PhantomData<T>);
 
     pub struct Normal;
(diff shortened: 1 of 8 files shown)


eravm-stable-interface/src/tracer_interface.rs
```diff
@@ -53,18 +53,19 @@ macro_rules! pub_struct {
 }
 
 pub mod opcodes {
+    use super::{CallingMode, ReturnType};
+    use std::marker::PhantomData;
+
     forall_simple_opcodes!(pub_struct);
-    pub struct FarCall<M: TypeLevelCallingMode>(M);
-    pub struct Ret<T: TypeLevelReturnType>(T);
```
joonazan, line 58: Does using PhantomData silence the warning about these not being used? If not, it is unnecessary.

src/callframe.rs
```diff
@@ -259,3 +259,28 @@ impl<T, W> Clone for Callframe<T, W> {
         }
     }
 }
+
+impl<T, W> PartialEq for Callframe<T, W> {
```
joonazan, line 263: Are the new PartialEq impls required for the test?

joonazan (approved):

eravm-stable-interface/src/tracer_interface.rs
slowli: Yep, it does; this is exactly why I've switched to it. Weirdly enough, these warnings aren't emitted in this repo, only when `vm2` is compiled in the Era codebase. Haven't figured out why yet.

PR #58, opened 2024-09-09
feat: Brush up repo for publishing

slowli:
# What ❔

Starts brushing up the repo for publishing:

- Better structures the repo directories.
- Renames crates in a uniform way + uses workspace dependencies.
- Structures imports using well-defined configuration.
- Reduces visibility of some items.
- Adds more crate metadata, e.g. licenses and readmes.

## Why ❔

Part of preparations for publishing the library crates.

## Checklist


crates/vm2/src/heap.rs
```diff
@@ -69,55 +69,8 @@ impl Heap {
...
-
-#[inline(always)]
-fn address_to_page_offset(address: u32) -> (usize, usize) {
-    let offset = address as usize;
-    (offset >> 12, offset & (HEAP_PAGE_SIZE - 1))
-}
-
-impl HeapInterface for Heap {
-    fn read_u256(&self, start_address: u32) -> U256 {
+    // TODO: reduce visibility once `multivm` uses `StateInterface` APIs
```
slowli, line 72: I plan to do this in a follow-up PR to keep the scope of this one reasonable.

crates/vm2-interface/src/lib.rs
```diff
@@ -23,9 +23,9 @@
 //! version that you publish and import it from the previous version instead.
 //!
 //! This is how you would add a new method to StateInterface and a new opcode.
+//!
 //! ```
-//! # use eravm_stable_interface as eravm_stable_interface_v1;
```
joonazan, line 27: I'd keep this renaming to make clearer that you are supposed to have the older version as a dependency.

crates/vm2/src/bitset.rs
```diff
@@ -1,5 +1,6 @@
+/// Bitset with 1,024 elements.
```
joonazan, line 1: 2^16 actually. Note that the array contains u64.

crates/vm2/src/heap.rs
joonazan: In theory, a lot of fields in state could be made private using this method as well.

joonazan (changes requested):
Fix the incorrect 1024 comment and I'll approve.

crates/vm2/src/heap.rs
slowli: Yep, that's the plan.

joonazan (approved):

PR #59, opened 2024-09-10
refactor: Brush up repo for publishing (pt. 2)

slowli:
# What ❔

Continues brushing up the repo.

- Makes VM state types private and somewhat extends `StateInterface` to compensate.
- Adds some doc comments (more to be added separately).

## Why ❔

Minimizes the API surface of the VM crate so that it's easier to maintain.

## Checklist

### DIFF
--- a/crates/vm2/src/addressing_modes.rs
+++ b/crates/vm2/src/addressing_modes.rs
@@ -1,3 +1,5 @@
+//! Addressing modes supported by the ZKsync Era VM.
+
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
 use enum_dispatch::enum_dispatch;
@@ -42,7 +44,7 @@ pub(crate) trait Destination {
 }
 
 /// The part of VM state that addressing modes need to operate on
-pub trait Addressable {
+pub(crate) trait Addressable {
     fn registers(&mut self) -> &mut [U256; 16];
     fn register_pointer_flags(&mut self) -> &mut u16;
 
@@ -73,7 +75,7 @@ impl<T: SourceWriter> SourceWriter for Option<T> {
 }
 
 #[enum_dispatch]
-pub trait DestinationWriter {
+pub(crate) trait DestinationWriter {
     fn write_destination(&self, args: &mut Arguments);
 }
 
@@ -109,7 +111,7 @@ impl Arguments {
     ) -> Self {
         // Make sure that these two can be packed into 8 bits without overlapping
         assert!(predicate as u8 & (0b11 << 6) == 0);
-        assert!(mode_requirements.0 & !(0b11) == 0);
+        assert!(mode_requirements.0 & !0b11 == 0);
 
         Self {
             source_registers: PackedRegisters(0),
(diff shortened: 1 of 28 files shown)


slowli:
I've added `U256` heap read / write methods to `StateInterface`. Otherwise, there's ~5% performance regression on some of batch benchmarks (e.g., with load test transactions). IMO, they improve DevEx as well.

crates/vm2/src/vm.rs
```diff
@@ -13,26 +13,24 @@ use crate::{
...
 }
 
+/// High-performance out-of-circuit ZKsync Era VM.
+#[derive(Debug)]
 pub struct VirtualMachine<T, W> {
-    pub world_diff: WorldDiff,
-
+    pub(crate) world_diff: WorldDiff,
     /// Storing the state in a separate struct is not just cosmetic.
     /// The state couldn't be passed to the world if it was inlined.
```
joonazan, line 32: I'd remove this comment. IIRC it is from a time when world was contained in the VM.

crates/vm2/src/addressing_modes.rs
```diff
@@ -427,12 +441,14 @@ impl Source for CodePage {
...
+/// Representation of one of 16 VM registers.
+#[derive(Debug, Clone, Copy)]
 pub struct Register(u8);
 
 impl Register {
-    pub fn new(n: u8) -> Self {
-        debug_assert!(n < 16);
+    /// Creates a register with the specified 0-based index.
+    pub const fn new(n: u8) -> Self {
+        assert!(n < 16, "Era VM has 16 registers");
```
joonazan, line 451:
```suggestion
        assert!(n < 16, "EraVM has 16 registers");
```

joonazan (commented):
You should consider checking if the new trait methods get properly inlined. Easiest way of doing this I know is to use record a perf profile and look for functions that are not inlined in firefox profiler's flamegraph.

crates/vm2/src/heap.rs
```diff
@@ -188,12 +182,18 @@ fn address_to_page_offset(address: u32) -> (usize, usize) {
     (offset >> 12, offset & (HEAP_PAGE_SIZE - 1))
 }
 
+#[derive(Debug, Clone, Copy)]
+enum HeapRollbackItem {
```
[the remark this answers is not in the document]
slowli: I'm for removing `write_byte`, although it *could* harm DevEx eventually 🙃 Regarding performance: not 100% sure. I've initially implemented history tracking separate bytes (https://github.com/matter-labs/vm2/pull/59/commits/77abefef2cf99dddc9193a2e7fe7277c84f9b687), and there was performance regression. I've implemented reading / writing `U256`s while still tracking separate bytes in history, and performance regression was still there. Then I've changed it to tracking `U256`s or bytes, and the regression was gone.

slowli:
> Easiest way of doing this I know is to use record a perf profile and look for functions that are not inlined in firefox profiler's flamegraph.

I've used cachegrind output produced by iai benches / `cg_annotate` for this purpose because `perf report` hangs for some reason locally; hopefully, it's an equivalent way of detecting inlining. It seems that the new methods *are* inlined.

PR #60, opened 2024-09-13
refactor: Brush up repo for publishing (pt. 3)

slowli:
# What ❔

- Documents all public items, applies the corresponding lint.
- Applies more clippy lints.
- Makes snapshot management private. Makes some more types / traits private as well (e.g., `PtrOp` trait).
- Exposes `Snapshot` and `StorageChange` types.
- Re-exports `zksync_vm2_interface` as the `interface` module instead of re-exporting its items.
- Removes potential UB during `offset_from` calls.

## Why ❔

Part of preparations for publishing the VM.

## Checklist


crates/vm2/src/callframe.rs
```diff
@@ -133,17 +135,29 @@ impl<T, W> Callframe<T, W> {
         })
     }
 
+    /// Gets a raw inferred program counter. This value can be garbage if the frame is on an invalid instruction or free panic.
+    #[allow(clippy::cast_possible_wrap)] // false positive: `Instruction` isn't that large
+    pub(crate) fn get_raw_pc(&self) -> isize {
```
slowli, line 140: I've tested performance relative to `offset_from` locally and there doesn't seem to be significant difference between the two.

crates/vm2/src/callframe.rs
```diff
@@ -133,17 +135,29 @@ impl<T, W> Callframe<T, W> {
...
+        // We cannot use `<*const _>::offset_from` because `self.pc` isn't guaranteed to be allocated within `self.program`
+        // (invalid instructions and free panics aren't).
+        let offset_in_bytes =
+            self.pc as isize - ptr::from_ref(self.program.instruction(0).unwrap()) as isize;
+        offset_in_bytes / mem::size_of::<Instruction<T, W>>() as isize
+    }
+
+    // FIXME: can overflow / underflow on invalid instruction / free panic
```
slowli, line 148: Are there guarantees that this method in this case? I don't think this is the case; e.g., the host can create a snapshot after VM execution has panicked.

crates/vm2/src/callframe.rs
```diff
@@ -133,17 +135,29 @@ impl<T, W> Callframe<T, W> {
...
 
     /// Sets the next instruction to execute to the instruction at the given index.
     /// If the index is out of bounds, the invalid instruction is used.
-    pub fn set_pc_from_u16(&mut self, index: u16) {
+    pub(crate) fn set_pc_from_u16(&mut self, index: u16) {
         self.pc = self
             .program
             .instruction(index)
-            .unwrap_or(invalid_instruction())
+            .unwrap_or_else(invalid_instruction);
```
slowli, line 160: I've initially [removed](https://github.com/matter-labs/vm2/pull/60/commits/9f72c5f95c0c751d56ca659c1bc8e66bdf381b39) `invalid_instruction` in favor of using an instruction inside a `Program` (it's already present as the last instruction in shorter programs); it also allows not to have a dummy `World` implementation for `Infallible`. However, this led to 5–10% regression on some iai benches.

crates/vm2/src/single_instruction_test/into_zk_evm.rs
```diff
@@ -333,7 +333,7 @@ impl VmWitnessTracer<8, EncodingModeProduction> for NoOracle {
         &mut self,
         _: u32,
         _: zk_evm::aux_structures::LogQuery,
-        _: zk_evm::aux_structures::PubdataCost,
```
slowli, line 336: I've initially enabled [`unused_qualifications` lint](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-qualifications), but it seems to be broken; it reports too many false positives to be usable.

crates/vm2/src/tests/mod.rs
slowli, line 1: Having tests in the library reduced compilation time and provides more opportunities for assertions etc. (although they aren't used for now).

slowli (commented):
- Can split the PR into parts if it makes it more reviewable.
- There are some debatable decisions (e.g., enabling pedantic clippy lints – I think it makes code better, but this is subjective).
- The built crate docs [are available](https://matter-labs.github.io/vm2/zksync_vm2_interface/).

crates/vm2/src/rollback.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::zero_sized_map_values)] // FIXME: why?
```
slowli, line 1: Dumb question: Is there a particular reason `BTreeMap<_, ()>` is used instead of `BTreeSet`?

slowli:
Here's the corresponding PR on the Era side: https://github.com/matter-labs/zksync-era/pull/2883. There's performance regression on some iai benches (order of 2–3%); will investigate its cause.

crates/vm2/src/lib.rs
```diff
@@ -65,6 +66,43 @@ pub trait StorageInterface {
     fn is_free_storage_slot(&self, contract: &H160, key: &U256) -> bool;
 }
 
+impl StorageInterface for Infallible {
+    fn read_storage(&mut self, _contract: H160, _key: U256) -> Option<U256> {
+        unreachable!("`Infallible` cannot be constructed")
```
joonazan, line 71:
```suggestion
        match *self {}
```

Why panic when you can invoke [Ex falso quodlibet](https://en.wikipedia.org/wiki/Principle_of_explosion) instead?

crates/vm2/src/callframe.rs
joonazan: You've highlighted an unsoundness bug that I introduced when I put all the per-instruction logic inside the instructions. The Invalid instruction code assumes that the handler isn't called but this is no longer true.

By putting an empty type as W of INVALID_INSTRUCTION, you actually introduce more unsoundness. The handler of `Instruction<(), Infallible>`, is a function that is impossible to call, as its World argument cannot be constructed. But then you cast it an instruction that can be called.

crates/vm2/src/rollback.rs
joonazan: I think RollbackableSet used to reuse RollbackableMap. Now that it is special-cased, it may as well use a set.

crates/vm2/src/callframe.rs
joonazan: We could store an invalid instruction in the VM like done with panic.

The perhaps nicer alternative is always calling panic and invalid instruction directly, skipping the dispatch loop.

crates/vm2-interface/src/state_interface.rs
```diff
@@ -1,93 +1,148 @@
...
     fn number_of_callframes(&self) -> usize;
+    /// Returns a mutable handle to a call frame with the specified index, where
     /// zero is the current frame, one is the frame before that etc.
     fn callframe(&mut self, n: usize) -> impl CallframeInterface + '_;
 
-    fn read_heap_byte(&self, heap: HeapId, index: u32) -> u8;
-    /// Reads an entire `U256` word in the big-endian order from the specified heap page / `index`
+    /// Reads a single byte from the specified heap page at the specified 0-based offset.
+    fn read_heap_byte(&self, heap: HeapId, offset: u32) -> u8;
+    /// Reads an entire `U256` word in the big-endian order from the specified heap page / `offset`
```
joonazan, line 20: I try to always write just "heap" rather than "memory page" because the latter's established meaning related to mapping physical memory to virtual memory.

```suggestion
    /// Reads an entire `U256` word in the big-endian order from the specified heap at `offset`
```

crates/vm2-interface/src/state_interface.rs
```diff
@@ -1,93 +1,148 @@
...
     fn set_flags(&mut self, flags: Flags);
 
+    /// Returns the currently set 0-based transaction number.
     fn transaction_number(&self) -> u16;
+    /// Sets the current transaction number.
     fn set_transaction_number(&mut self, value: u16);
 
+    /// Returns the value of the context register.
     fn context_u128_register(&self) -> u128;
+    /// Sets the value of the context register.
```
joonazan, line 39: This comment adds no information, which is why I don't like to have mandatory comments. I would explain the difference between the context register and the context field of the callframe here but I don't remember what the difference is.

crates/vm2-interface/src/state_interface.rs
```diff
@@ -1,93 +1,148 @@
...
+/// Public interface of an EraVM call frame.
 pub trait CallframeInterface {
+    /// Address of the storage context associated with this frame. For delegate calls, this address is inherited from the calling contract;
+    /// otherwise, it's the same as [`Self::code_address()`].
     fn address(&self) -> H160;
+    /// Sets the address of the executing contract.
     fn set_address(&mut self, address: H160);
+    /// Address of the contract being executed.
     fn code_address(&self) -> H160;
+    /// Sets the address of the contract being executed.
```
joonazan, line 82: This should document if changing the address causes new code to be loaded.

crates/vm2-interface/src/state_interface.rs
```diff
@@ -1,93 +1,148 @@
...
     fn set_code_address(&mut self, address: H160);
+    /// Address of the calling contract. Respects delegate and mimic calls.
     fn caller(&self) -> H160;
+    /// Sets the address of the calling contract
     fn set_caller(&mut self, address: H160);
 
-    /// During panic and arbitrary code execution this returns None.
+    /// Returns the current program counter (i.e., 0-based index of the instruction being executed).
+    /// During panic and arbitrary code execution this returns `None`.
```
joonazan, line 90: Arbitrary code execution was removed from the interface for now. Panic maybe doesn't move the pc if you implement my suggestion.

crates/vm2-interface/src/state_interface.rs
```diff
@@ -1,93 +1,148 @@
...
+    /// Checks whether the call is static.
     fn is_static(&self) -> bool;
+    /// Checks whether the call is executed in the kernel mode.
     fn is_kernel(&self) -> bool;
 
+    /// Returns the remaining amount of gas.
     fn gas(&self) -> u32;
+    /// Sets the remaining amount of gas.
     fn set_gas(&mut self, new_gas: u32);
+    /// Additional
```
joonazan, line 110: ... gas provided for the duration of this callframe?

crates/vm2-interface/src/state_interface.rs
```diff
@@ -1,93 +1,148 @@
...
     fn set_program_counter(&mut self, value: u16);
 
+    /// Returns a 0-based index of the starting exception handler instruction.
     fn exception_handler(&self) -> u16;
+    /// Sets the exception handler pointer.
     fn set_exception_handler(&mut self, value: u16);
 
+    /// Checks whether the call is static.
     fn is_static(&self) -> bool;
+    /// Checks whether the call is executed in the kernel mode.
```
joonazan, line 103:
```suggestion
    /// Checks whether the call is executed in kernel mode.
```

crates/vm2-interface/src/state_interface.rs
```diff
@@ -1,93 +1,148 @@
...
-    /// During panic and arbitrary code execution this returns None.
+    /// Returns the current program counter (i.e., 0-based index of the instruction being executed).
+    /// During panic and arbitrary code execution this returns `None`.
     fn program_counter(&self) -> Option<u16>;
-
+    /// Sets the program counter.
     /// The VM will execute an invalid instruction if you jump out of the program.
     fn set_program_counter(&mut self, value: u16);
 
+    /// Returns a 0-based index of the starting exception handler instruction.
```
joonazan, line 96: This and the comment for set use very different terminology. How about something like "Returns the program counter that the parent frame continue from if this frame fails"

crates/vm2-interface/src/state_interface.rs
```diff
@@ -1,93 +1,148 @@
...
     fn is_kernel(&self) -> bool;
 
+    /// Returns the remaining amount of gas.
     fn gas(&self) -> u32;
+    /// Sets the remaining amount of gas.
     fn set_gas(&mut self, new_gas: u32);
+    /// Additional
     fn stipend(&self) -> u32;
 
+    /// Returns the context register for this call.
```
joonazan, line 113: The context_u128 of a callframe is not a register

crates/vm2-interface/src/state_interface.rs
```diff
@@ -1,93 +1,148 @@
...
     fn read_stack(&self, index: u16) -> (U256, bool);
+    /// Sets the value and pointer flag for the specified stack slot.
     fn write_stack(&mut self, index: u16, value: U256, is_pointer: bool);
 
+    /// Returns the stack pointer.
     fn stack_pointer(&self) -> u16;
+    /// Sets the stack pointer.
     fn set_stack_pointer(&mut self, value: u16);
 
+    /// Returns ID of the main heap page used in this call.
```
joonazan, line 131: These three comments also say "heap page"

joonazan: Actually, even read_code_page could be named read_contract_code

crates/vm2-interface/src/state_interface.rs
```diff
@@ -109,33 +164,49 @@ impl HeapId {
...
 
-#[derive(Debug)]
+/// L2-to-L1 log emitted by EraVM.
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub struct L2ToL1Log {
+    /// Log key.
     pub key: U256,
+    /// Log value.
     pub value: U256,
+    /// Is this a service log?
```
joonazan, line 198: I actually don't understand where the is_service terminology comes from. At least with events it is sometimes used interchangeably with is_first.

crates/vm2/src/addressing_modes.rs
```diff
@@ -133,7 +136,7 @@ impl Arguments {
             1..=4 => panic!("Reserved gas cost values overlap with actual gas costs"),
             x => {
                 if x > u8::MAX as u32 {
```
joonazan, line 138: The cleaner way that satisfies clippy is to use `.try_into()`.

crates/vm2/src/callframe.rs
```diff
@@ -54,7 +56,7 @@ pub(crate) struct NearCallFrame {
     world_before_this_frame: Snapshot,
 }
 
-impl<T, W> Callframe<T, W> {
+impl<T: Tracer, W: World<T>> Callframe<T, W> {
```
joonazan, line 59: These bounds are unnecessary.

crates/vm2/src/callframe.rs
joonazan: Well spotted! Looks like I forgot to read why offset_from is unsafe.

crates/vm2/src/decode.rs
```diff
@@ -1,44 +1,31 @@
...
-        .chain(std::iter::once(if raw.len() >= 1 << 16 {
-            jump_to_beginning()
-        } else {
-            Instruction::from_invalid()
-        }))
-        .collect()
-}
-
-fn unimplemented_instruction<T, W>(variant: Opcode) -> Instruction<T, W> {
+fn unimplemented_instruction<T: Tracer, W: World<T>>(variant: Opcode) -> Instruction<T, W> {
```
joonazan, line 28: Again, I'd leave out trait bounds whenever possible. Makes future refactorings easier.

crates/vm2/src/decommit.rs
```diff
@@ -98,7 +98,8 @@ impl WorldDiff {
         let code = world.decommit_code(code_hash);
         if is_new {
             // Decommitter can process two words per cycle
```
joonazan, line 100: Move this comment one line down.

joonazan: Also, I wonder if there is a clearer way of writing "divide by 64, rounding up"

crates/vm2/src/fat_pointer.rs
```diff
@@ -1,34 +1,41 @@
...
-        unsafe { &mut *(value as *mut U256).cast() }
+        unsafe { &mut *ptr::from_mut(value).cast() }
     }
 }
 
 #[cfg(target_endian = "little")]
 impl From<U256> for FatPointer {
     fn from(value: U256) -> Self {
         unsafe {
-            let ptr: *const FatPointer = (&value as *const U256).cast();
```
joonazan, line 25: I think the cleanest code for this is `transmute(value.low_u128())`

crates/vm2/src/heap.rs
```diff
@@ -217,20 +217,21 @@ impl Heaps {
     }
 
     fn allocate_inner(&mut self, memory: &[u8]) -> HeapId {
-        let id = HeapId::from_u32_unchecked(self.heaps.len() as u32);
-        self.heaps
+        let id = u32::try_from(self.inner.len()).expect("heap ID overflow");
```
joonazan, line 220: Allocating 2^32 heaps with 2^32 gas would be quite a feat. Also, we'd have a memory use problem.

Especially once we start reusing heap ids, this is pretty much impossible. Do we start reusing heap ids, tough? It might be a bit confusing for code that keeps a HeapId for too long but that shouldn't be happening much.

crates/vm2/src/heap.rs
```diff
@@ -184,7 +184,7 @@ fn address_to_page_offset(address: u32) -> (usize, usize) {
 
 #[derive(Debug, Clone)]
 pub(crate) struct Heaps {
-    heaps: Vec<Heap>,
+    inner: Vec<Heap>,
```
joonazan, line 187: IMO the old name was more descriptive.

crates/vm2/src/instruction.rs
```diff
@@ -27,28 +28,15 @@ pub(crate) enum ExecutionStatus {
     Stopped(ExecutionEnd),
 }
 
+/// End of a VM execution returned from [`VirtualMachine::run()`].
```
joonazan, line 31:
```suggestion
/// VM stop reason; returned from [`VirtualMachine::run()`].
```

crates/vm2/src/instruction_handlers/binop.rs
```diff
@@ -13,22 +18,21 @@ use crate::{
...
-fn binop<
+fn binop<T, W, Op, In1, Out, const SWAP: bool, const SET_FLAGS: bool>(
+    vm: &mut VirtualMachine<T, W>,
+    world: &mut W,
+    tracer: &mut T,
+) -> ExecutionStatus
+where
     T: Tracer,
-    W,
+    W: World<T>,
```
joonazan, line 31:
```suggestion
    W,
```

crates/vm2/src/instruction_handlers/binop.rs
```diff
@@ -104,7 +108,7 @@ impl Binop for Xor {
 impl Binop for ShiftLeft {
     #[inline(always)]
     fn perform(a: &U256, b: &U256) -> (U256, (), Flags) {
-        let result = *a << b.low_u32() as u8;
+        let result = *a << (b.low_u32() & 0xff);
```
joonazan, line 111:
```suggestion
        let result = *a << (b.low_u32() % 256);
```
Maybe this would make it clearer that the mask isn't arbitrary.

crates/vm2/src/instruction_handlers/binop.rs
```diff
@@ -231,18 +235,19 @@ macro_rules! from_binop {
...
         ) -> Self {
-            Self::from_binop::<$binop>(src1, src2, out, out2, arguments, swap, set_flags)
+            Self::from_binop::<$binop>(src1, src2, out, &out2, arguments, swap, set_flags)
         }
     };
 }
 
-impl<T: Tracer, W> Instruction<T, W> {
+/// Instructions for binary operations.
+impl<T: Tracer, W: World<T>> Instruction<T, W> {
```
joonazan, line 244:
```suggestion
impl<T: Tracer, W> Instruction<T, W> {
```

crates/vm2/src/instruction_handlers/common.rs
```diff
@@ -1,36 +1,46 @@
 use zksync_vm2_interface::{opcodes, OpcodeType, Tracer};
 
 use super::ret::free_panic;
-use crate::{addressing_modes::Arguments, instruction::ExecutionStatus, VirtualMachine};
+use crate::{addressing_modes::Arguments, instruction::ExecutionStatus, VirtualMachine, World};
 
 #[inline(always)]
-pub(crate) fn boilerplate<Opcode: OpcodeType, T: Tracer, W>(
+pub(crate) fn boilerplate<Opcode, T, W>(
```
joonazan, line 7: IMO where clauses are good when types don't fit onto one line or when they are necessary. Do you or clippy disagree?

crates/vm2/src/instruction_handlers/context.rs
```diff
@@ -8,64 +8,75 @@ use zksync_vm2_interface::{
...
+
+fn context<T, W, Op>(
     vm: &mut VirtualMachine<T, W>,
     world: &mut W,
     tracer: &mut T,
-) -> ExecutionStatus {
+) -> ExecutionStatus
+where
+    T: Tracer,
+    W: World<T>,
```
joonazan, line 29: I'll stop commenting about unnecessary bounds to limit the spam.

crates/vm2/src/instruction_handlers/context.rs
```diff
@@ -8,64 +8,75 @@ use zksync_vm2_interface::{
...
 
 impl ContextOp for SP {
-    fn get<T, W>(state: &State<T, W>) -> U256 {
+    fn get<T: Tracer, W: World<T>>(state: &State<T, W>) -> U256 {
         state.current_frame.sp.into()
     }
 }
 
-fn context_meta<T: Tracer, W>(
+#[allow(clippy::cast_sign_loss)] // intentional
```
joonazan, line 78: I believe you can put this annotation before the offending line to make it more obvious.

crates/vm2/src/instruction_handlers/far_call.rs
```diff
@@ -182,15 +185,15 @@ pub(crate) fn get_far_call_arguments(abi: U256) -> FarCallABI {
...
+pub(crate) fn get_far_call_calldata<T: Tracer, W: World<T>>(
     raw_abi: U256,
     is_pointer: bool,
     vm: &mut VirtualMachine<T, W>,
     already_failed: bool,
 ) -> Option<FatPointer> {
     let mut pointer = FatPointer::from(raw_abi);
 
-    match FatPointerSource::from_abi((raw_abi.0[3] >> 32) as u8) {
+    match FatPointerSource::from_abi((raw_abi.0[3] >> 32) & 0xff) {
```
joonazan, line 196: I prefer the version that clippy dislikes here. Maybe that lint should be turned off.

crates/vm2/src/instruction_handlers/pointer.rs
```diff
@@ -102,11 +114,31 @@ impl PtrOp for PointerShrink {
...
+
+/// Pointer-related instructions.
+impl<T: Tracer, W: World<T>> Instruction<T, W> {
+    from_ptr_op!(from_pointer_add<PointerAdd>);
+    from_ptr_op!(from_pointer_sub<PointerSub>);
+    from_ptr_op!(from_pointer_pack<PointerPack>);
+    from_ptr_op!(from_pointer_shrink<PointerShrink>);
 
-impl<T: Tracer, W> Instruction<T, W> {
     #[inline(always)]
```
joonazan, line 140: The `Instruction::from_*` family sometimes has `#[inline(always)]` and sometimes doesn't. I guess it doesn't hurt because they are only used once but the compiler should be able to figure that out, so I'd say remove all the annotations.

crates/vm2/src/instruction_handlers/ret.rs
```diff
@@ -1,23 +1,34 @@
+use std::convert::Infallible;
```
joonazan, line 1: Infallible is easy to read when used in a Result but I'd use `!` or `enum Void {}` elsewhere.

crates/vm2/src/mode_requirements.rs
```diff
@@ -1,17 +1,20 @@
-#[derive(Debug)]
+/// Requirements for the VM execution mode that can be placed by instructions.
```
joonazan, line 1:
```suggestion
/// VM execution mode requirements (kernel only, not in static call) that can be placed on instructions.
```

crates/vm2/src/predication.rs
```diff
@@ -4,43 +4,46 @@ const GT_BIT: u8 = 1 << 2;
...
+    /// Execute the associated instruction if the "equal" execution flag is set.
     IfEQ = EQ_BIT,
+    /// Execute the associated instruction if the "less than" execution flag is set.
     IfLT = LT_BIT,
+    /// Execute the associated instruction if either of "greater than" or "equal" execution flags are set.
     IfGE = GT_BIT | EQ_BIT,
+    /// Execute the associated instruction if either of "less than" or "equal" execution flags are set.
     IfLE = LT_BIT | EQ_BIT,
+    /// Execute the associated instruction if the "equal" execution flag is not set.
     IfNotEQ = EQ_BIT << 4 | ALWAYS_BIT,
```
joonazan, line 34: This is cute but just having `NOT_EQ_BIT = 1 << 4` and handling the special case might be faster and clearer.

crates/vm2/src/program.rs
```diff
@@ -75,3 +116,36 @@ impl<T, W> PartialEq for Program<T, W> {
             && Arc::ptr_eq(&self.instructions, &other.instructions)
     }
 }
+
+/// "Jump to start" instruction placed at the end of programs exceeding `1 << 16` instructions.
```
joonazan, line 120:
```suggestion
/// Wraparound instruction placed at the end of programs exceeding `1 << 16` instructions to simulate the 16-bit program counter overflowing.
/// Does not invoke tracers because it is an implementation detail, not an actual instruction.
```

crates/vm2/src/single_instruction_test/into_zk_evm.rs
```diff
@@ -91,7 +94,7 @@ pub struct MockMemory {
 // One arbitrary heap value is not enough for zk_evm
```
joonazan, line 94: This comment is correct but seems to be in the wrong place.

crates/vm2/src/single_instruction_test/into_zk_evm.rs
```diff
@@ -91,7 +94,7 @@ pub struct MockMemory {
...
 // because it reads two U256s to read one U256.
```
joonazan, line 95: This comment is correct but seems to be in the wrong place.

crates/vm2/src/single_instruction_test/into_zk_evm.rs
```diff
@@ -121,7 +124,7 @@ impl Memory for MockMemory {
     ) -> zk_evm::aux_structures::MemoryQuery {
         match query.location.memory_type {
             MemoryType::Stack => {
-                let slot = query.location.index.0 as u16;
+                let slot = u16::try_from(query.location.index.0).unwrap();
```
joonazan, line 127: I believe this must wrap, not panic.

crates/vm2/src/single_instruction_test/mod.rs
```diff
@@ -5,6 +5,8 @@
 //!
 //! The same kind of mocking in applied to stack memory, the program, the world and callstack.
```
joonazan, line 6:
```suggestion
//! The same kind of mocking is applied to stack memory, the program, the world and callstack.
```

crates/vm2/src/stack.rs
```diff
@@ -19,6 +19,7 @@ const NUMBER_OF_DIRTY_AREAS: usize = 64;
 const DIRTY_AREA_SIZE: usize = (1 << 16) / NUMBER_OF_DIRTY_AREAS;
 
 impl Stack {
+    #[allow(clippy::cast_ptr_alignment)] // aligned per `Stack` layout
```
joonazan, line 22: Maybe switching Stack to Self satisfies the lint?

crates/vm2/src/vm.rs
```diff
@@ -128,44 +134,66 @@ impl<T: Tracer, W> VirtualMachine<T, W> {
...
+
+        self.snapshot = Some(VmSnapshot {
             world_snapshot: self.world_diff.external_snapshot(),
             state_snapshot: self.state.snapshot(),
-        }
+        });
     }
 
-    /// Returns the VM to the state it was in when the snapshot was created.
+    /// Returns the VM to the state it was in when [`Self::make_snapshot()`] was called. If there is
```
joonazan, line 156: ... one?

crates/vm2/src/world_diff.rs
```diff
@@ -290,7 +277,7 @@ impl WorldDiff {
         }
     }
 
-    pub(crate) fn rollback(&mut self, snapshot: Snapshot) {
+    pub(crate) fn rollback(&mut self, snapshot: &Snapshot) {
```
joonazan, line 280: The Snapshot should be consumed. Rolling back to it multiple times isn't intended use.

joonazan: Same with ExternalSnapshot

crates/vm2-interface/src/state_interface.rs
slowli: Mandatory comments looks prettier in the generated docs. Otherwise, I agree that having meaningful method names is sometimes enough 😄

crates/vm2/src/instruction_handlers/common.rs
slowli: No strong opinion from either me or clippy, so I'll revert it back.

crates/vm2/src/instruction_handlers/ret.rs
slowli: `!` is unstable and `Infallible` is [its stable replacement](https://doc.rust-lang.org/std/convert/enum.Infallible.html#future-compatibility). `enum Void {}` would technically work, but I'd rather not create additional types.

crates/vm2/src/stack.rs
slowli: No, it doesn't. `alloc_zeroed` returns `*mut u8`, which has lower alignment requirements compared to `Layout`. clippy isn't smart enough to figure out that the actual alignment depends on the arg supplied to `alloc_zeroed` (I mean, it's very difficult / impossible to figure out in the general case).

crates/vm2/src/single_instruction_test/into_zk_evm.rs
slowli: Where would you suggest to place it?

crates/vm2/src/decommit.rs
slowli: There is [`div_ceil`](https://doc.rust-lang.org/core/primitive.u64.html#method.div_ceil).

crates/vm2/src/callframe.rs
slowli: > The perhaps nicer alternative is always calling panic and invalid instruction directly, skipping the dispatch loop.

I think this can work for panic; AFAICT, setting `pc` to point to panic is always invoked in an instruction handler and is followed by the handler termination. But in the case of invalid instructions, could you explain how it would work e.g., in this place (i.e., outside an instruction handler)?

crates/vm2/src/predication.rs
slowli: Sorry, I don't understand this suggestion; e.g., what is the special case here? You can implement it separately 🙃

crates/vm2/src/heap.rs
slowli: I'd consider this a theoretical concern, but for correctness, using `try_from` + `expect` is preferable. IIUC, it doesn't influence performance that much.

crates/vm2/src/callframe.rs
slowli: I've added more details to the comment and "downgraded" it to TODO.

crates/vm2/src/callframe.rs
joonazan: Ok, I'll fix the unsoundness separately then.

crates/vm2/src/decommit.rs
joonazan: div_ceil looks perfect. Communicates the intention well, so I'd replace (x + 63) / 64 and (x + 1) / 2 with this method.

crates/vm2/src/predication.rs
joonazan: (not important) All the other cases are handled by `predicate & flags != 0` but that can't handle IfNotEq, so that one is handled by having the next four bits mean flags that we don't want.

crates/vm2/src/single_instruction_test/into_zk_evm.rs
joonazan: The function below: partially_overlapping...

Looks like the comment is from a draft version where there were two values in the struct. But I ended up working around the issue with that function instead.

joonazan (dismissed):
Looks good once the remaining few comments are implemented.

I will fix the invalid instruction unsoundness in a separate PR.

crates/vm2/src/decommit.rs
slowli: Yeah, I did just that 😄

joonazan (approved):

PR #61, opened 2024-09-18
fix: invalid instruction unsoundness

joonazan:
Fixes a case of unsoundness.

The fix involves a 100% safe Rust way to write generic static variables, so I use it for panic as well.
This may buy us some performance.


slowli (approved):
Nice 👍

PR #64, opened 2024-09-20
fix: fuzz.sh

joonazan:
The path in the file was incorrect.


slowli (approved):
Crap, I haven't noticed this file when I was refactoring the repository structure.

PR #66, opened 2024-09-20
ci: Fix secret used for publishing

slowli:
# What ❔

Uses a dedicated secret for publishing workflows. 

## Why ❔

This is necessary because of `GITHUB_TOKEN` limitations.

## Checklist


.github/workflows/pr-title.yml
popzxc, line 1: If the concern is duplicated workflow, probably it's better to replace old one with this one. Old one uses [deprecated action](https://github.com/aslafy-z/conventional-pr-title-action) and doesn't add a comment explaining how to fix the PR title.
Up to you, however, no strong opinion.

popzxc (dismissed):

.github/workflows/pr-title.yml
slowli: I've added a hack to the new workflow instead: a new step that adds the necessary status check. It seems to work, although it's difficult to tell with 100% of certainty because the old PR title workflow somehow continues to run despite being removed by the PR 🙃

joonazan (approved):

PR #69, opened 2024-09-24
fix: some methods operating on wrong near call

joonazan:
Targeting near calls in the tracer API was very buggy and the code for it was rather convoluted. Both issues are fixed and regression tests are added.

Also fixes an issue where setting the pc would always set the current pc instead of the pc of the targeted frame.

### DIFF
--- a/crates/vm2/src/tracing.rs
+++ b/crates/vm2/src/tracing.rs
@@ -310,7 +310,11 @@ impl<T: Tracer, W> CallframeInterface for CallframeWrapper<'_, T, W> {
     }
 
     fn set_program_counter(&mut self, value: u16) {
-        self.frame.set_pc_from_u16(value);
+        if let Some(call) = self.near_call_on_top_mut() {
+            call.previous_frame_pc = value;
+        } else {
+            self.frame.set_pc_from_u16(value);
+        }
     }
 
     fn exception_handler(&self) -> u16 {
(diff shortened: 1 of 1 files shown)


crates/vm2/src/tracing.rs
```diff
@@ -335,11 +335,10 @@ impl<T, W> CallframeWrapper<'_, T, W> {
         if self.frame.near_calls.is_empty() || self.near_call == Some(0) {
```
slowli, line 335: I admit I haven't thoroughly looked at this piece of code before, so: how is `near_call` defined? In the `current_frame()` impl, it seems to align with `frame.near_calls` indexing, but in `callframe()`, the opposite indexing is used; e.g., `callframe(0)` will translate to `near_call == None` or `Some(0)`, depending on whether there are near calls in the frame. AFAIU, the logic below and `exception_handler` methods assume reverse indexing. IMO, it would be easier to wrap a hand around direct indexing; in this case, the near call on top can be retrieved using `self.frame.near_calls.get(next_index)`, where `next_index = self.near_call.map_or(0, |idx| idx + 1)`.

joonazan: Yes, it is probably better to do the index reversal earlier.
Also, even with this fix CallTracer doesn't behave correctly, so I think I need to check for other issues, too.

slowli (approved):

PR #73, opened 2024-09-26
fix!: Revise VM storage interface

slowli:
# What ❔

Replaces storage slot data returned by `StorageInterface` with a dedicated type (essentially, `(U256, bool)`). 

## Why ❔

Currently, `StorageInterface` reads storage slots as `Option<U256>` where `None` corresponds to initial writes. This loses information during sandboxed execution. Indeed, there may be slots which have non-zero value, but still are initial writes (i.e., these slots were first written to during previous blocks / transactions in the sandboxed batch).

## Checklist

### DIFF
--- a/crates/vm2/src/lib.rs
+++ b/crates/vm2/src/lib.rs
@@ -49,15 +49,32 @@ mod tracing;
 mod vm;
 mod world_diff;
 
+/// Storage slot information returned from [`StorageInterface::read_storage()`].
+#[derive(Debug, Clone, Copy)]
+pub struct StorageSlot {
+    /// Value of the storage slot.
+    pub value: U256,
+    /// Whether a write to the slot would be considered an initial write. This influences refunds.
+    pub is_write_initial: bool,
+}
+
+impl StorageSlot {
+    /// Represents an empty storage slot.
+    pub const EMPTY: Self = Self {
+        value: U256([0; 4]),
+        is_write_initial: true,
+    };
+}
+
 /// VM storage access operations.
 pub trait StorageInterface {
     /// Reads the specified slot from the storage.
     ///
     /// There is no write counterpart; [`WorldDiff::get_storage_changes()`] gives a list of all storage changes.
-    fn read_storage(&mut self, contract: H160, key: U256) -> Option<U256>;
+    fn read_storage(&mut self, contract: H160, key: U256) -> StorageSlot;
 
     /// Computes the cost of writing a storage slot.
-    fn cost_of_writing_storage(&mut self, initial_value: Option<U256>, new_value: U256) -> u32;
+    fn cost_of_writing_storage(&mut self, initial_slot: StorageSlot, new_value: U256) -> u32;
 
     /// Returns if the storage slot is free both in terms of gas and pubdata.
     fn is_free_storage_slot(&self, contract: &H160, key: &U256) -> bool;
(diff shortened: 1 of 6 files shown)


crates/vm2/src/world_diff.rs
```diff
@@ -372,102 +378,198 @@ mod tests {
...
     proptest! {
         #[test]
-        fn test_storage_changes(
-            initial_values in arbitrary_storage_changes(),
+        fn storage_changes_work_as_expected(
+            initial_values in arbitrary_initial_storage(),
             first_changes in arbitrary_storage_changes(),
             second_changes in arbitrary_storage_changes(),
         ) {
-            let storage_initial_values = initial_values
```
slowli, line 383: Moved this out of the macro for reuse, and because the Rust formatter is notoriously bad at formatting code inside macros (it mostly ignores badly formatted code).

crates/vm2/src/world_diff.rs
```diff
@@ -369,105 +375,195 @@ const COLD_WRITE_AFTER_WARM_READ_REFUND: u32 = STORAGE_ACCESS_COLD_READ_COST;
...
+                .prop_map(|(contract, key)| (H160::from(contract), U256::from(key))),
+            any::<([u8; 32], bool)>().prop_map(|(value, is_write_initial)| StorageSlot {
+                value: U256::from(value),
+                is_write_initial,
+            }),
+            0..=MAX_ITEMS,
+        )
+    }
+
+    fn constrained_initial_storage() -> impl Strategy<Value = BTreeMap<(H160, U256), StorageSlot>> {
```
slowli, line 527: The unconstrained strategy fails to catch at least some simple bugs, e.g. setting `is_initial: initial.value.is_zero()` in `get_storage_changes_after()`.

joonazan:
For maximum efficiency with Postgres there'd have to have one method returning (U256, bool) and another returning just U256. I was planning to do that optimization at some point but it is quite complex, so I think it might not be worth it.

I would write the integration of this so that RocksDB/hashmap cache is only queried once per method call. That would help performance a lot, at least according some benchmark flamegraphs.

slowli:
> For maximum efficiency with Postgres there'd have to have one method returning (U256, bool) and another returning just U256.

That's what I've implemented 🙃 I agree that it looks like the most efficient approach. Regarding changes in the Era codebase, that's a separate topic that we can approach iteratively.

crates/vm2/src/lib.rs
```diff
@@ -49,15 +49,39 @@ mod tracing;
...
+    /// Same as [`Self::read_storage()`], but doesn't request the initialness flag for the read slot.
+    ///
+    /// The default implementation uses `read_storage()`.
+    fn read_storage_value(&mut self, contract: H160, key: U256) -> U256 {
+        self.read_storage(contract, key).value
+    }
 
     /// Computes the cost of writing a storage slot.
-    fn cost_of_writing_storage(&mut self, initial_value: Option<U256>, new_value: U256) -> u32;
+    fn cost_of_writing_storage(&mut self, initial_slot: StorageSlot, new_value: U256) -> u32;
```
joonazan, line 84: What do you think about merging this method and `read_storage`? It would require moving the initial value cache outside the VM but we already have such a cache there.

joonazan (approved):

crates/vm2/src/lib.rs
slowli: Maybe, but it's worth tackling separately IMO; the current data flow looks reasonable to me as is.

PR #74, opened 2024-09-26
fix: don't grow heap unless it is paid for

joonazan:
Undoes the tangles that had to be added to replicate two zk_evm bugs.

zk_evm and circuits are finally merging bugfixes.


Cargo.toml
```diff
@@ -30,9 +30,9 @@ primitive-types = "0.12.1"
 proptest = "1.4"
 
 # "Internal" dependencies
-zkevm_opcode_defs = "0.150.0"
-zk_evm_abstractions = "0.150.0"
-zk_evm = "0.150.5"
+zkevm_opcode_defs = "0.151.3"
+zk_evm_abstractions = "0.151.3"
+zk_evm = "0.151.3"
```
slowli, line 35: Along with updating the deps, you can resolve [a FIXME in the `callframe` module](https://github.com/matter-labs/vm2/blob/56cc787b578ef689b74e779c646aa70945e2c7f2/crates/vm2/src/callframe.rs#L16-L17) by importing `NEW_EVM_FRAME_MEMORY_STIPEND` from `zkevm_opcode_defs::system_params`.

crates/vm2/src/instruction_handlers/far_call.rs
```diff
@@ -184,40 +181,29 @@ pub(crate) fn get_far_call_calldata<T, W>(
...
-                        grow_heap::<_, _, AuxHeap>(&mut vm.state, bound).ok()?;
-                        pointer.memory_page = vm.state.current_frame.aux_heap;
-                    }
+            if is_pointer || pointer.offset != 0 {
+                return None;
+            }
+            let bound = pointer
+                .start
+                .checked_add(pointer.length)
+                .unwrap_or(u32::MAX);
```
slowli, line 197: Am I missing something, or would `saturating_add` work here as well? If it would, it looks more idiomatic.

crates/vm2/src/instruction_handlers/heap_access.rs
```diff
@@ -153,11 +148,9 @@ pub(crate) fn grow_heap<T, W, H: HeapFromState>(
     state: &mut State<T, W>,
     new_bound: u32,
 ) -> Result<(), ()> {
-    let already_paid = H::get_heap_size(state);
-    if *already_paid < new_bound {
-        let to_pay = new_bound - *already_paid;
-        *already_paid = new_bound;
+    if let Some(to_pay) = new_bound.checked_sub(*H::get_heap_size(state)) {
         state.use_gas(to_pay)?;
+        *H::get_heap_size(state) = new_bound;
```
slowli, line 153: Should this go before or after the `use_gas()` call? (The ordering matters since the call short-circuits on an error.) AFAIU, the previous logic changed the heap bound before the call – was it a part of the buggy behavior?

slowli (commented):
- I think cargo-deny action is outdated, so that's the reason 3 corresponding workflows fail. E.g., Era [uses](https://github.com/matter-labs/zksync-era/blob/d44ceca9f109f08f9d23f147535e54b89875c8ae/.github/workflows/cargo-license.yaml#L8) a newer commit (`8d73959fce1cdc8989f23fdf03bec6ae6a6576e`, version 2.0.7).
- As for [another divergence during fuzzing](https://github.com/matter-labs/vm2/actions/runs/13900563475/job/38891026065?pr=74), is it clear what it's caused by?

crates/vm2/src/instruction_handlers/heap_access.rs
joonazan: It is important that it comes after so size only grows if it can be paid for.

crates/vm2/src/instruction_handlers/far_call.rs
joonazan: True, though no longer relevant as in the final version I had to differentiate between overflow happening or not.

joonazan:
Fuzzing indicates no more divergences, though I silenced an issue where zk_evm sometimes passes calldata even to failed far calls. That behaviour doesn't really affect anything but I don't want to replicate it.

joonazan:
Actually, the fuzzer found one more divergence, though it might be unrelated. Will investigate tomorrow.

slowli (approved):

PR #76, opened 2024-09-30
perf: more lightweight second gas limit

joonazan:
The validation gas limit requires being able to limit the amount of gas that is processed. The original implementation had to scan through the whole callstack on every single instruction. This fixes that so that the scan is done at most once per call.


slowli (commented):
I may be missing something, but I think this logic doesn't always work. Suppose we have `gas_limit = 1000` with the execution in the bootloader with 10000 gas left; i.e., initially  `minimum_gas = 10000 - 1000 = 9000`. Suppose the bootloader calls into another contract passing 500 gas to it. Then, since `500 < minimum_gas`, `minimum_gas` will be updated to `500.saturating_sub(1000 - spent) = 0` (assuming `spent` is sufficiently small). Then, `self.state.current_frame.gas < minimum_gas` will always be false (including after the called contract returns), so gas limiting won't work.

PR #78, opened 2024-10-28
feat: allow tracers to stop execution

joonazan:
Tracers can now request to stop the VM in `Tracer::after_instruction`

Makes validation tracer implementation nicer.

### DIFF
--- a/crates/vm2-interface/src/lib.rs
+++ b/crates/vm2-interface/src/lib.rs
@@ -105,6 +105,7 @@ impl ExecutionStatus {
     /// Combines [`ExecutionStatus`] values from two sources, choosing the most severe one.
     /// So if one tracer wants to suspend but the other wants to panic, the VM will panic.
     #[must_use]
+    #[inline(always)]
     pub fn merge(self, other: Self) -> Self {
         use ExecutionStatus::{Running, Stopped};
         match (self, other) {
--- a/crates/vm2-interface/src/lib.rs
+++ b/crates/vm2-interface/src/lib.rs
@@ -91,3 +91,41 @@ pub use self::{state_interface::*, tracer_interface::*};
 
 mod state_interface;
 mod tracer_interface;
+
+/// Returned from Tracer::after_instruction to indicate whether the VM should continue execution.
+#[derive(Debug)]
+pub enum ExecutionStatus {
+    /// Continue execution.
+    Running,
+    /// Stop or suspend execution.
+    Stopped(ExecutionEnd),
+}
+
+impl ExecutionStatus {
+    /// Combines ExecutionStatus values from two sources, choosing the most severe one.
+    /// So if one tracer wants to suspend but the other wants to panic, the VM will panic.
+    pub fn merge(self, other: Self) -> Self {
+        match (self, other) {
+            (
+                ExecutionStatus::Stopped(ExecutionEnd::SuspendedOnHook(_)),
+                ExecutionStatus::Stopped(end),
+            ) => ExecutionStatus::Stopped(end),
+            (ExecutionStatus::Stopped(end), _) => ExecutionStatus::Stopped(end),
+            (_, ExecutionStatus::Stopped(end)) => ExecutionStatus::Stopped(end),
+            _ => ExecutionStatus::Running,
+        }
+    }
+}
+
+/// VM stop reason and return value.
+#[derive(Debug, PartialEq)]
+pub enum ExecutionEnd {
+    /// The executed program has finished and returned the specified data.
+    ProgramFinished(Vec<u8>),
+    /// The executed program has reverted returning the specified data.
+    Reverted(Vec<u8>),
+    /// The executed program has panicked.
+    Panicked,
+    /// Returned when the bootloader writes to the heap location specified by [`hook_address`](crate::Settings.hook_address).
+    SuspendedOnHook(u32),
+}
--- a/crates/vm2-interface/src/lib.rs
+++ b/crates/vm2-interface/src/lib.rs
@@ -28,6 +28,7 @@
 //! # use zksync_vm2_interface as zksync_vm2_interface_v1;
 //! use zksync_vm2_interface_v1::{
 //!     StateInterface as StateInterfaceV1, GlobalStateInterface as GlobalStateInterfaceV1, Tracer as TracerV1, opcodes::NearCall,
+//!     ShouldStop,
 //! };
 //!
 //! trait StateInterface: StateInterfaceV1 {
@@ -60,7 +61,9 @@
 //!
 //! trait Tracer {
 //!     fn before_instruction<OP: OpcodeType, S: GlobalStateInterface>(&mut self, state: &mut S) {}
-//!     fn after_instruction<OP: OpcodeType, S: GlobalStateInterface>(&mut self, state: &mut S) {}
+//!     fn after_instruction<OP: OpcodeType, S: GlobalStateInterface>(&mut self, state: &mut S) -> ShouldStop {
+//!         ShouldStop::Continue
+//!     }
 //! }
 //!
 //! impl<T: TracerV1> Tracer for T {
@@ -73,7 +76,9 @@
 //!             }
 //!         }
 //!     }
-//!     fn after_instruction<OP: OpcodeType, S: GlobalStateInterface>(&mut self, state: &mut S) {}
+//!     fn after_instruction<OP: OpcodeType, S: GlobalStateInterface>(&mut self, state: &mut S) -> ShouldStop {
+//!         todo!()
+//!     }
 //! }
 //!
 //! // Now you can use the new features by implementing TracerV2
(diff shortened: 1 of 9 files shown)


crates/vm2-interface/src/lib.rs
```diff
@@ -91,3 +91,40 @@ pub use self::{state_interface::*, tracer_interface::*};
...
 mod state_interface;
 mod tracer_interface;
+
+/// Returned from [`Tracer::after_instruction`] to indicate whether the VM should continue execution.
+#[derive(Debug)]
+pub enum ExecutionStatus {
+    /// Continue execution.
+    Running,
+    /// Stop or suspend execution.
+    Stopped(ExecutionEnd),
```
slowli, line 101: I don't quite like that a tracer can stop execution with `ProgramFinished` or `Reverted` outcomes; it looks weird / unnatural. Not quite sure about `Panicked` either. Maybe, it'd be sufficient to just allow tracers suspend execution on a hook?

crates/vm2-interface/src/tracer_interface.rs
```diff
@@ -254,17 +254,26 @@ impl<T: opcodes::TypeLevelReturnType> OpcodeType for opcodes::Ret<T> {
...
+    /// The tracer's return value takes precedence over the VM but only if it is at least as severe.
+    /// For example, if the VM wants to stop and the tracer wants to suspend, the VM will still stop.
     ///
     /// The default implementation does nothing.
-    fn after_instruction<OP: OpcodeType, S: GlobalStateInterface>(&mut self, state: &mut S) {
+    #[must_use]
+    fn after_instruction<OP: OpcodeType, S: GlobalStateInterface>(
+        &mut self,
+        state: &mut S,
+    ) -> ExecutionStatus {
```
slowli, line 274: Maybe a dumb suggestion: What if this method is left unchanged, and instead there is a mutating method in `StateInterface` allowing to set execution end (or just suspend execution on a hook, see the other line comment) after the current instruction? (Internally, it would store the hook somewhere and check it in the boilerplate code, for example.) Would this break some invariants or tank performance? I know that we don't care about backward compatibility at this point, but it would be an example of a backward-compatible change (vs breaking changes in the current design).

joonazan: There is a way to do it without performance hit: point the PC at an instruction that stops execution. However, that is much more complex than this.

If you think that returning an ExecutionStatus from after_instruction is a hassle, I'd rather make `after_instruction_that_can_diverge` where the default implementation calls `after_instruction`. Slightly more convenient but also harder to understand.

crates/vm2-interface/src/lib.rs
joonazan: I completely disagree on this one. When the validation tracer ends execution prematurely, that is certainly a panic. When the validation tracer ends execution because the validation step ended, that is ProgramFinished. (I currently don't do the latter in the tracer but I would have if I did the hooks via tracer)

crates/vm2-interface/src/tracer_interface.rs
slowli: Just to clarify: Does having an additional field for the tracer return status actually harm performance significantly? What order of regression are we talking about?

> Slightly more convenient but also harder to understand.

Not worth it, IMO. If we did care about backward compatibility, this would be an OK-ish option, but rn it looks like a half-measure.

crates/vm2-interface/src/lib.rs
slowli: In my mind, all "normal" ways to end execution have certain implicit invariants associated with them, such as having only the root call frame (e.g., unwinding in case of panics). Having these invariants breakable by tracers looks like a suboptimal design decision; hooks are good in this regard because they explicitly possible to encounter at any point in execution and allow to the host to decide how to proceed. A downside of this approach is that the host needs to be aware of hooks that can be set by tracers, but specifically for validation it's not an issue (it's easy to agree that a certain hook has "execution stopped by tracer" semantics).

crates/vm2-interface/src/tracer_interface.rs
joonazan: > Does having an additional field for the tracer return status actually harm performance significantly?

A struct field is harmless. The problem is having to check that field to decide if we want to continue or stop. The compiler can't do global reasoning, so it will assume the field could change at any time and produces code that checks it at the end of every instruction.

I haven't benchmarked this particular thing but having an extra function call that determines whether to continue tanked performance by 20-50% and this is very similar.

crates/vm2-interface/src/lib.rs
joonazan: One advantage that I can see is that the VM can know from the hook number which tracer is responsible for the hook. I can also see that stopping while there are stack frames is unusual and dangerous.

On the other hand I have thought that suspending should not be allowed because it is completely useless for (external) tracers, as the multivm will just continue running immediately.

joonazan: I think the only use for this feature is "I want to stop the VM early to save on computation". The validation tracer would also like to signal if it stopped because of success or failure but that can be done via the tracer.

Maybe there should be another VM end state: StoppedByTracer?

crates/vm2-interface/src/tracer_interface.rs
slowli: I've tested performance locally using [this branch](https://github.com/matter-labs/vm2/compare/aov/status-in-vm-state), comparing against the `main` branch in Era, which uses commit df5bec3d04d64d434f9b0ccb285ba4681008f7b3. There's ~1% performance regression on the instruction benchmark on all bytecodes other than `decode_shl_sub` (~5% regression) and `event_spam` (~3% regression):

```text
[√] init/fast
├ Instructions           4288946       -14286 (-0.33%)
└ Est. cycles            6065233         -206 (-0.00%)
[√] access_memory/fast
├ Instructions         712286449     -2846496 (-0.40%)
└ Est. cycles         1081996604      +960296 (+0.09%)
[√] call_far/fast
├ Instructions         414705094     -2800354 (-0.67%)
└ Est. cycles          530854193     -4626916 (-0.86%)
[√] decode_shl_sub/fast
├ Instructions         480051957    +17770031 (+3.84%)
└ Est. cycles          670545005    +32025441 (+5.02%)
[√] deploy_simple_contract/fast
├ Instructions          12691330        +4082 (+0.03%)
└ Est. cycles           22498200       +53958 (+0.24%)
[√] event_spam/fast
├ Instructions         530783134     +6147136 (+1.17%)
└ Est. cycles          827640827    +22601422 (+2.81%)
[√] finish_eventful_frames/fast
├ Instructions         738161974      -898719 (-0.12%)
└ Est. cycles         1158831668     +7756344 (+0.67%)
[√] heap_read_write/fast
├ Instructions         562962994     +1630928 (+0.29%)
└ Est. cycles          799787757     +6235806 (+0.79%)
[√] slot_hash_collision/fast
├ Instructions         536992272     +1576282 (+0.29%)
└ Est. cycles          777353068    +11034097 (+1.44%)
[√] write_and_decode/fast
├ Instructions         608095839     +3157674 (+0.52%)
└ Est. cycles          857472348     +9515405 (+1.12%)
```

Oneshot benchmark results mostly look small, other than `slot_hash_collision` (~10% regression), `decode_shl_sub` (~5%) and `write_and_decode` (~3%):

```text
                             main                             branch
fast/access_memory           [37.663 ms 37.907 ms 38.237 ms]  [37.460 ms 37.524 ms 37.602 ms]
fast/call_far                [31.226 ms 31.302 ms 31.384 ms]  [31.298 ms 31.485 ms 31.781 ms]
fast/decode_shl_sub          [20.973 ms 21.055 ms 21.161 ms]  [22.024 ms 22.095 ms 22.159 ms]
fast/deploy_simple_contract  [1.4368 ms 1.5502 ms 1.6506 ms]  [1.4294 ms 1.4390 ms 1.4459 ms]
fast/event_spam              [40.702 ms 40.830 ms 41.009 ms]  [40.844 ms 40.960 ms 41.064 ms]
fast/finish_eventful_frames  [68.124 ms 68.935 ms 70.298 ms]  [66.591 ms 66.901 ms 67.243 ms]
fast/heap_read_write         [30.268 ms 30.318 ms 30.367 ms]  [30.767 ms 30.872 ms 30.956 ms]
fast/slot_hash_collision     [26.223 ms 26.303 ms 26.379 ms]  [28.899 ms 28.972 ms 29.081 ms]
fast/write_and_decode        [27.764 ms 27.849 ms 27.969 ms]  [28.739 ms 28.805 ms 28.885 ms]
```

Batch benchmarks (1,000 txs / batch) have small regression:

```text
                             main                             branch
deploy_simple_contract       [338.44 ms 343.26 ms 348.84 ms]  [337.52 ms 339.68 ms 341.84 ms]
load_test                    [372.71 ms 374.52 ms 376.43 ms]  [380.68 ms 381.29 ms 381.90 ms]
load_test_realistic          [319.64 ms 320.02 ms 320.58 ms]  [321.14 ms 326.86 ms 330.88 ms]
load_test_heavy              [670.01 ms 672.60 ms 675.04 ms]  [676.75 ms 678.39 ms 680.05 ms]
transfer                     [270.67 ms 271.42 ms 272.35 ms]  [273.64 ms 273.88 ms 274.34 ms]
```

Subjectively, ~5% regression (on synthetic / worst-case inputs!) would be acceptable in exchange for cleaner code. (E.g., I *think* boilerplate wrappers can be simplified further in this approach.) WDYT? Maybe, my refactoring is not equivalent or insufficient to support tracers?

joonazan: Performance is better than I thought, though you have also changed the dispatch loop. The alternative is zero-cost, though.

Switching to tail calls will remove the branch from the loop but won't be able to remove the branch caused by this formulation. Perhaps a more representative benchmark would be to leave the loop as-is and add the check to boilerplate.

ValidationTracer has one stop (out of gas) that has to be detected in before_instruction, so that is a case that devolves to code that resembles this solution but only in case of panic.

I don't really see not having to put a return value into a tracer method as worth it.

joonazan: Tried the loop optimization. It makes things worse, so it probably improves performance for you because it avoids mem::take

crates/vm2-interface/src/lib.rs
slowli: From the correctness perspective, an additional add state would be optimal. If it doesn't regress performance, I'd go with it.

crates/vm2-interface/src/tracer_interface.rs
slowli: > Perhaps a more representative benchmark would be to leave the loop as-is and add the check to boilerplate.

Dumb question: what would it represent? Naturally, I've changed the dispatch loop to retain a single execution stop check per loop iteration (so that there's lesser performance regression).

My concern with the proposed design is that the approach conceptually looks unmaintainable in the long term, since adding stuff to be returned by tracers will break their interface every time. This is very vague, of course; so if you don't see it as a concern, let's go with your approach.

joonazan: I think I know the future: tracers will want to stop and communicate their stop reason, which they can do via some other channel. Tracers may want to suspend in order to implement hooks but at least for now the hardcoded hooks are just better.

crates/vm2-interface/src/lib.rs
joonazan: I will implement that

slowli (approved):

PR #79, opened 2025-01-20
feat: Allow customizing precompiles logic

slowli:
# What ❔

Allows customizing the precompiles logic by extending the `World` trait.

## Why ❔

- Allows caching `ecrecover` calls (= signature verification) with the default account abstraction.
- Precompiles is the only entry point for crypto backends, so this allows theoretically customizing a backend (not implemented in this PR).

## Checklist


crates/vm2/src/precompiles/legacy.rs
```diff
@@ -0,0 +1,417 @@
...
+        RecoveryId,
+        Digest,
+        R,
+        S,
+        Key,
+    }
+
+    #[test]
+    fn ecrecover_precompile_works() {
+        let mut rng = rand::thread_rng();
```
joonazan, line 235: Consider making proptests instead of non-deterministic tests.

crates/vm2/src/precompiles/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+use zksync_vm2_interface::CycleStats;
+
+pub use self::legacy::LegacyPrecompiles;
+use crate::heap::Heap;
+
+mod legacy;
+
+/// Provides access to the input memory for a precompile call.
+#[derive(Debug, Clone)]
+pub struct PrecompileMemoryReader<'a, const IN_WORDS: bool = false> {
```
joonazan, line 17: The type parameter IN_WORDS doesn't seem to be used for anything. Is it necessary even though there is already `assume_offset_in_words` for multiplying the offset and length?

Nit: I'd make an alternative constructor instead.

joonazan (approved):
This is a very good idea! I found integrating the precompiles directly somewhat dirty.

crates/vm2/src/precompiles/mod.rs
slowli: The param is used for type safety. E.g., it's impossible to chain `assume_offset_in_words()` calls; if a `PrecompileMemoryReader` is used as an arg, it's always clear whether length / offset is measured in bytes or in words, etc. In [the initial commit](https://github.com/matter-labs/vm2/pull/79/commits/e4302c3a443be97b8c20daf5811abca253e994f0), I've had more streamlined precompile implementations [that show this more clearly](https://github.com/matter-labs/vm2/blob/e4302c3a443be97b8c20daf5811abca253e994f0/crates/vm2/src/precompiles.rs) (decided to remove them for now since I'm not 100% sure they correspond 1-to-1 to the legacy implementations).

> I'd make an alternative constructor instead.

This would be problematic because `PrecompileMemoryReader` instances are constructed inside the VM code, not in `Precompiles` implementations; so it would necessitate the VM code being aware which precompiles use byte / word offsets.

joonazan (approved):

PR #81, opened 2025-01-20
fix: failing far call tracing

joonazan:
Fixes tracing so that a failed far call reports a far call and then a panic to tracers instead of reporting the panic in the middle of the far call.

### DIFF
--- a/crates/vm2/src/instruction_handlers/far_call.rs
+++ b/crates/vm2/src/instruction_handlers/far_call.rs
@@ -12,7 +12,6 @@ use super::{
     common::full_boilerplate,
     heap_access::grow_heap,
     monomorphization::{match_boolean, monomorphize, parameterize},
-    ret::{panic_from_failed_far_call, RETURN_COST},
     AuxHeap, Heap,
 };
 use crate::{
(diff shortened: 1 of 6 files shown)


crates/vm2/src/instruction_handlers/far_call.rs
```diff
@@ -107,10 +106,9 @@ where
         vm.state.current_frame.gas -= normally_passed_gas;
         let new_frame_gas = normally_passed_gas + mandated_gas;
 
-        let Some((calldata, program, is_evm_interpreter)) = failing_part else {
-            vm.state.current_frame.gas += new_frame_gas.saturating_sub(RETURN_COST);
-            return panic_from_failed_far_call(vm, world, tracer, exception_handler);
-        };
+        // A far call pushes a new frame and returns from it in the next instruction if it panics.
+        let (calldata, program, is_evm_interpreter) =
+            failing_part.unwrap_or((U256::zero().into(), Program::new_panicking(), false));
```
slowli, line 111: Nit: It makes sense to use `unwrap_or_else` here to allocate a `Program` lazily.

slowli (dismissed):

PR #82, opened 2025-01-24
fix!: Remove EVM emulator gas stipend

slowli:
# What ❔

Removes the EVM emulator gas stipend and replaces it with a memory stipend.

## Why ❔

Aligns with the changes in the legacy VM.

## Checklist


slowli:
- I've tested this code on `multivm` unit tests (with the legacy VM revision updated, of course), so there's *some* insurance that it works correctly. OTOH, I had to remove the `stipend` test because it no longer made sense. Would it be possible to replace it somehow?
- Breaking changes could theoretically be removed / moved to another PR.

crates/vm2/src/single_instruction_test/callframe.rs
```diff
@@ -42,7 +42,6 @@ impl<'a, T: Tracer, W: World<T>> Arbitrary<'a> for Callframe<T, W> {
             sp: u.arbitrary()?,
             // It is assumed that it is always possible to add the stipend
             gas: u.int_in_range(0..=u32::MAX - EVM_SIMULATOR_STIPEND)?,
```
joonazan, line 44: I think this no longer needs to be constrained, as nothing adds gas anymore. The constant can be deleted as well.

crates/vm2/src/single_instruction_test/state_to_zk_evm.rs
```diff
@@ -86,7 +86,7 @@ fn vm2_frame_to_zk_evm_frames<T, W>(
         heap_bound: frame.heap_size,
         aux_heap_bound: frame.aux_heap_size,
         total_pubdata_spent: PubdataCost(0),
-        stipend: frame.stipend,
```
joonazan, line 89: They forgot to remove the field? Or does the zk_evm dependency need updating?

crates/vm2/src/tracing.rs
```diff
@@ -458,7 +454,7 @@ mod test {
             1000,
             crate::Settings {
                 default_aa_code_hash: [0; 32],
-                evm_interpreter_code_hash: [0; 32],
```
joonazan, line 461: This pervasive renaming seems somewhat unnecessary. The other renamings did improve clarity.

joonazan (changes requested):
Good apart from a few little things.
Removing the test is fine because the system replacing stipends isn't nearly as dangerous.

crates/vm2/src/single_instruction_test/state_to_zk_evm.rs
slowli: I've postponed updating the `zk_evm` dependency until it's released, yes.

joonazan (approved):

PR #84, opened 2025-03-05
ci: fix PR title workflow by removing unnecessary check

antonbaliasnikov:
# What ❔

* [x] Fix PR title workflow by removing the unnecessary check
* [x] Fix clippy issues 

## Why ❔

Workflow is not working properly for external collaborators if `github-status-action-v2` is in use, it's unnecessary and should be removed.

## Checklist

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -8,7 +8,7 @@ on:
       - synchronize
 
 jobs:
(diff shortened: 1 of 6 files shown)


joonazan (approved):
Apparently the low precedence of bitwise or is no longer considered common knowledge

PR #87, opened 2025-04-04
feat: Elliptic curve precompiles

joonazan:
Adds elliptic curve precompiles support.

### DIFF
--- a/crates/vm2/src/precompiles/legacy.rs
+++ b/crates/vm2/src/precompiles/legacy.rs
@@ -440,7 +440,7 @@ mod tests {
             LegacyPrecompiles.call_precompile(SECP256R1_VERIFY_PRECOMPILE_ADDRESS, memory, 0);
 
         prop_assert_eq!(output.len, 2);
-        let [is_ok, is_verified, _is_verified] = output.buffer;
+        let [is_ok, is_verified, _] = output.buffer;
         if mutation.is_none() {
             prop_assert_eq!(is_ok, U256::one());
             prop_assert_eq!(is_verified, U256::one());
--- a/crates/vm2/src/precompiles/legacy.rs
+++ b/crates/vm2/src/precompiles/legacy.rs
@@ -73,9 +73,16 @@ impl Memory for LegacyIo<'_> {
     ) -> MemoryQuery {
         let start_word = query.location.index.0;
         if query.rw_flag {
-            assert!(start_word < 2, "standard precompiles never write >2 words");
-            self.output.buffer[start_word as usize] = query.value;
-            self.output.len = self.output.len.max(start_word + 1);
+            assert!(start_word < 3, "standard precompiles never write >3 words");
+            // start_word == 2 only for EcAdd and EcMul precompiles
+            if start_word == 2 {
+                self.output.buffer[0] = self.output.buffer[1];
+                self.output.buffer[1] = query.value;
+                self.output.len = self.output.len.max(start_word);
+            } else {
+                self.output.buffer[start_word as usize] = query.value;
+                self.output.len = self.output.len.max(start_word + 1);
+            }
         } else {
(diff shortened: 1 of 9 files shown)


Raid5594 (dismissed):

crates/vm2/src/precompiles/mod.rs
```diff
@@ -89,17 +89,17 @@ impl From<U256> for PrecompileOutput {
...
 {
     fn from(value: [U256; N]) -> Self {
         let mut buffer = [U256::zero(); 3];
         buffer[..N].copy_from_slice(&value[..N]);
-        
+
         Self {
             buffer,
-            len: N as u32,
+            len: u32::try_from(N).expect("Not a valid length"),
```
joonazan, line 102: Nit: you can set N to u32, which maybe lets you use `.into()` to infallibly convert it to usize.

crates/vm2/src/precompiles/mod.rs
```diff
@@ -82,18 +82,24 @@ impl PrecompileOutput {
...
             cycle_stats: None,
         }
     }
 }
 
-impl From<[U256; 2]> for PrecompileOutput {
-    fn from(value: [U256; 2]) -> Self {
+impl<const N: usize> From<[U256; N]> for PrecompileOutput
+where
+    [U256; N]: Default,
```
slowli, line 94: This conversion is too general and doesn't make sense for `N > 3`. If you want to avoid copy-paste please define a macro via macro_rules taking the array length as a param and invoke it for 2 and 3 (and optionally, 1):

```rust
macro_rules! impl_from_array_for_precompile_output {
    ($n:tt) => {
        impl From<[U256; $n]> for PrecompileOutput {
            fn from(value: [U256; $n]) -> Self {
                todo!()
            }
        }
    };
}

impl_from_array_for_precompile_output!(2);
impl_from_array_for_precompile_output!(3);
```

crates/vm2/src/precompiles/legacy.rs
```diff
@@ -311,7 +335,7 @@ mod tests {
 
         prop_assert_eq!(output.len, 2);
         let expected_address = key_to_address(signing_key.verifying_key());
-        let [is_success, address] = output.buffer;
+        let [is_success, address, _address] = output.buffer;
```
slowli, line 338: Nit: Please use `_` for the third word here and below; it doesn't have address / "is_verified" semantics.

slowli (changes requested):

crates/vm2/src/precompiles/mod.rs
```diff
@@ -82,23 +82,34 @@ impl PrecompileOutput {
...
+macro_rules! impl_from_array_for_precompile_output {
+    ($n:tt) => {
+        impl From<[U256; $n]> for PrecompileOutput {
+            fn from(value: [U256; $n]) -> Self {
+                let mut buffer = [U256::zero(); 3];
+                buffer[..$n].copy_from_slice(&value[..$n]);
+
+                Self {
+                    buffer,
+                    len: u32::try_from($n).expect("Not a valid length"),
```
slowli, line 101: Nit: AFAIU, you can just use `len: $n` here.

slowli (approved):
I know this is unrelated to the PR subject, but `cargo-deny` jobs (which are required CI checks) fail and it would be somewhat easy to fix them by adding the "Unicode-3.0" license to [the list of allowed licenses](https://github.com/matter-labs/vm2/blob/831104788ad8250158a6e45b85ebd0a7157510d9/deny.toml#L15) (could make sense to replace "Unicode-DFS-2016", since [it's not encountered](https://github.com/matter-labs/vm2/actions/runs/14512626475/job/40714585813?pr=87#step:4:127)).

slowli (approved):

PR #95, opened 2026-02-25
feat: Extend functionality & fix regressions

popzxc:
This PR implements missing functionality required to verify execution via [tee_verifier](https://github.com/matter-labs/zksync-era/blob/main/core/lib/tee_verifier/src/lib.rs).
This is needed to prove EraVM in Airbender.
Additionally, this PR fixes several discovered divergencies vs `zk_evm`.

This is NOT a complete feature parity with `zk_evm` -- some things are explicitly not supported. For example, we don't generate full metadata in logs -- we generate just enough
data to verify execution, not to generate witness.
Similarly, we don't mimic behavior for non-zero shards (except for illegal calls to non-zero shards) as we don't care about them.
The goal was to come up with a minimum set of changes to make it provable and remove _realistic_ risks wrt divergencies.

Disclaimer: changes in this PR were majorly written by an LLM, but with a lot of manual interventions.

Since there is a lot of stuff going on in here, I've tried to add pointers that should make review easier.
Additionally, below is an LLM-generated overview of added features:

## Static UMA read/write support

vm2 now decodes and executes StaticMemoryRead / StaticMemoryWrite instead of treating them as missing functionality. This closes a core opcode-surface gap with zk_evm for static memory operations.

- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113)

## Static-memory isolation and precompile page-0 remapping

vm2 keeps static UMA memory isolated while preserving zk_evm's precompile ABI sentinel semantics for page `0`. This avoids accidental cross-contamination between static memory and regular heap behavior.

- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/uma.rs#L109-L113](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/uma.rs#L109-L113)
  - [crates/zk_evm/src/opcodes/execution/log.rs#L315-L335](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L315-L335)

## Nonzero-shard far-call side-effect ordering

vm2 now follows zk_evm-style ordering by failing nonzero-shard far calls before touching deployer storage in the unsupported-shard path. This removes an unnecessary side effect that previously created parity risk.

- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/far_call.rs#L110-L164](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/far_call.rs#L110-L164)

## Non-kernel returndata forwarding restriction for older pages

vm2 enforces non-kernel unidirectional returndata forwarding so callees cannot forward pointers to older pages. This aligns a subtle but important safety rule with zk_evm.

- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/ret.rs#L58-L75](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/ret.rs#L58-L75)

## Precompile extra-ergs out-of-gas behavior alignment

On insufficient extra ergs, vm2 now follows the non-panicking continuation behavior expected by zk_evm precompile execution semantics. This avoids turning this path into an unnecessary panic divergence.

- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/log.rs#L88-L134](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L88-L134)
  - [crates/zk_evm/src/opcodes/execution/log.rs#L294-L305](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L294-L305)

## Non-fresh decommit page reuse and lazy materialization

vm2 no longer assumes non-fresh decommits always have a preassigned page and no longer panics the host when that mapping is missing. It now reuses mapped pages when present and lazily materializes them when needed.

- zk_evm reference:
  - [crates/zk_evm/src/reference_impls/decommitter.rs#L61-L69](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/reference_impls/decommitter.rs#L61-L69)

## Decommit page lifetime protection across frame pop and rollback

vm2 preserves decommit pages that must remain globally reusable even when nested frames are popped or VM snapshots roll back. This prevents accidental deallocation of pages that zk_evm semantics treat as reusable history.

- zk_evm reference:
  - [crates/zk_evm/src/reference_impls/decommitter.rs#L61-L69](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/reference_impls/decommitter.rs#L61-L69)
  - [crates/zk_evm/src/vm_state/helpers.rs#L187-L223](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/vm_state/helpers.rs#L187-L223)
  - [crates/zk_evm/src/vm_state/helpers.rs#L232-L241](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/vm_state/helpers.rs#L232-L241)

## Storage log query stream (reads, writes, rollback markers)

vm2 now records a full storage access stream instead of write-only net diffs, including rollback markers on reverted writes. This is the required shape for parity-oriented dedup and witness-related consumers.

- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/log.rs#L143-L178](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L143-L178)
  - [crates/zk_evm/src/opcodes/execution/log.rs#L199-L236](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L199-L236)
  - [crates/zk_evm/src/vm_state/helpers.rs#L148-L176](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/vm_state/helpers.rs#L148-L176)

## Callstack-full masking before instruction execution

vm2 performs callstack-full masking in common instruction boilerplate, matching the high-level zk_evm control-flow behavior. The current depth estimator is intentionally approximate and documented with TODO because practical saturation is considered extremely unlikely.

- zk_evm reference:
  - [crates/zk_evm/src/vm_state/cycle.rs#L199-L223](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/vm_state/cycle.rs#L199-L223)
  - [crates/zk_evm/src/vm_state/mod.rs#L105-L107](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/vm_state/mod.rs#L105-L107)
  - [crates/zk_evm/src/vm_state/execution_stack.rs#L117-L119](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/vm_state/execution_stack.rs#L117-L119)

## Known divergencies that are NOT implemented:

We don't implement them since they don't seem to affect the execution for our use cases.

### Event emission scope parity

Status: Not implemented

vm2 still gates event emission by a specific event-writer address, while zk_evm's event path is broader in kernel context execution. This remains an accepted divergence in the current branch scope.

- vm2 current behavior:
  - `crates/vm2/src/instruction_handlers/event.rs:17-29`
- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/log.rs#L252-L289](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L252-L289)

### Full shard metadata and shard behavior parity

Status: Not implemented

vm2 continues to treat shard support as limited scope, with multiple paths hardcoding shard-related values. Full shard semantics and metadata parity with zk_evm are intentionally deferred.

- vm2 current behavior:
  - `crates/vm2/src/instruction_handlers/context.rs:82-98`
  - `crates/vm2/src/instruction_handlers/far_call.rs:160-177`
- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/far_call.rs#L94-L122](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/far_call.rs#L94-L122)
  - [crates/zk_evm/src/opcodes/execution/log.rs#L64-L68](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L64-L68)
  - [crates/zk_evm/src/opcodes/execution/log.rs#L276-L281](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L276-L281)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ members = [
     "crates/vm2-interface",
     "crates/vm2",
     # Testing crates
-    # "tests/afl-fuzz"
+    "tests/afl-fuzz"
 ]
 resolver = "2"
 
@@ -30,9 +30,9 @@ primitive-types = "0.12.1"
 proptest = "1.4"
 
 # "Internal" dependencies
-zkevm_opcode_defs = { path = "../zksync-protocol/crates/zkevm_opcode_defs" }
-zk_evm_abstractions = { path = "../zksync-protocol/crates/zk_evm_abstractions" }
-zk_evm = { path = "../zksync-protocol/crates/zk_evm" }
+zkevm_opcode_defs = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
+zk_evm_abstractions = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
+zk_evm = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
 
 # Dependencies within the workspace
 zksync_vm2_interface = { version = "=0.5.0", path = "crates/vm2-interface" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ members = [
     "crates/vm2-interface",
     "crates/vm2",
     # Testing crates
-    "tests/afl-fuzz"
+    # "tests/afl-fuzz"
 ]
 resolver = "2"
 
@@ -30,9 +30,9 @@ primitive-types = "0.12.1"
 proptest = "1.4"
 
 # "Internal" dependencies
-zkevm_opcode_defs = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
-zk_evm_abstractions = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
-zk_evm = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
+zkevm_opcode_defs = { path = "../zksync-protocol/crates/zkevm_opcode_defs" }
+zk_evm_abstractions = { path = "../zksync-protocol/crates/zk_evm_abstractions" }
+zk_evm = { path = "../zksync-protocol/crates/zk_evm" }
 
 # Dependencies within the workspace
 zksync_vm2_interface = { version = "=0.5.0", path = "crates/vm2-interface" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,9 +30,9 @@ primitive-types = "0.12.1"
 proptest = "1.4"
 
 # "Internal" dependencies
-zkevm_opcode_defs = { version ="0.153.0", path = "../zksync-protocol/crates/zkevm_opcode_defs" }
-zk_evm_abstractions = { version ="0.153.0", path = "../zksync-protocol/crates/zk_evm_abstractions" }
-zk_evm = { version ="0.153.0", path = "../zksync-protocol/crates/zk_evm" }
+zkevm_opcode_defs = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
+zk_evm_abstractions = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
+zk_evm = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
 
 # Dependencies within the workspace
 zksync_vm2_interface = { version = "=0.5.0", path = "crates/vm2-interface" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,9 +30,9 @@ primitive-types = "0.12.1"
 proptest = "1.4"
 
 # "Internal" dependencies
-zkevm_opcode_defs = "0.153.0"
-zk_evm_abstractions = "0.153.0"
-zk_evm = "0.153.0"
+zkevm_opcode_defs = { version ="0.153.0", path = "../zksync-protocol/crates/zkevm_opcode_defs" }
+zk_evm_abstractions = { version ="0.153.0", path = "../zksync-protocol/crates/zk_evm_abstractions" }
+zk_evm = { version ="0.153.0", path = "../zksync-protocol/crates/zk_evm" }
 
 # Dependencies within the workspace
 zksync_vm2_interface = { version = "=0.5.0", path = "crates/vm2-interface" }
--- a/crates/vm2/src/callframe.rs
+++ b/crates/vm2/src/callframe.rs
@@ -144,14 +144,17 @@ impl<T: Tracer, W: World<T>> Callframe<T, W> {
     pub(crate) fn get_raw_pc(&self) -> usize {
         // We cannot use `<*const _>::offset_from` because `self.pc` isn't guaranteed to be allocated within `self.program`
         // (invalid instructions and free panics aren't).
+        // We cannot use `isize` either, because it causes troubles when compiled to `RISC-V`, so we are
+        // returning the value that is clearly out of scope.
         let offset_in_bytes = (self.pc as usize)
-            .wrapping_sub(ptr::from_ref(self.program.instruction(0).unwrap()) as usize);
+            .checked_sub(ptr::from_ref(self.program.instruction(0).unwrap()) as usize)
+            .unwrap_or(usize::MAX);
         offset_in_bytes / mem::size_of::<Instruction<T, W>>()
     }
 
     // TODO: can overflow / underflow after an invalid instruction or free panic. Ordinarily, this will lead to VM termination (for an invalid instruction)
     //   or the callframe getting popped (for a panic), but it's still technically possible to invoke this method afterwards in certain cases (e.g., in the bootloader callframe).
-    #[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
+    #[allow(clippy::cast_possible_truncation)]
     pub(crate) fn get_pc_as_u16(&self) -> u16 {
         self.get_raw_pc() as u16
     }
--- a/crates/vm2/src/callframe.rs
+++ b/crates/vm2/src/callframe.rs
@@ -144,9 +144,9 @@ impl<T: Tracer, W: World<T>> Callframe<T, W> {
     pub(crate) fn get_raw_pc(&self) -> usize {
         // We cannot use `<*const _>::offset_from` because `self.pc` isn't guaranteed to be allocated within `self.program`
         // (invalid instructions and free panics aren't).
-        let offset_in_bytes =
-            self.pc as usize - ptr::from_ref(self.program.instruction(0).unwrap()) as usize;
-        offset_in_bytes / mem::size_of::<Instruction<T, W>>() as usize
+        let offset_in_bytes = (self.pc as usize)
+            .wrapping_sub(ptr::from_ref(self.program.instruction(0).unwrap()) as usize);
+        offset_in_bytes / mem::size_of::<Instruction<T, W>>()
     }
 
     // TODO: can overflow / underflow after an invalid instruction or free panic. Ordinarily, this will lead to VM termination (for an invalid instruction)
--- a/crates/vm2/src/callframe.rs
+++ b/crates/vm2/src/callframe.rs
@@ -141,12 +141,12 @@ impl<T: Tracer, W: World<T>> Callframe<T, W> {
 
     /// Gets a raw inferred program counter. This value can be garbage if the frame is on an invalid instruction or free panic.
     #[allow(clippy::cast_possible_wrap)] // false positive: `Instruction` isn't that large
-    pub(crate) fn get_raw_pc(&self) -> isize {
+    pub(crate) fn get_raw_pc(&self) -> usize {
         // We cannot use `<*const _>::offset_from` because `self.pc` isn't guaranteed to be allocated within `self.program`
         // (invalid instructions and free panics aren't).
         let offset_in_bytes =
-            self.pc as isize - ptr::from_ref(self.program.instruction(0).unwrap()) as isize;
-        offset_in_bytes / mem::size_of::<Instruction<T, W>>() as isize
+            self.pc as usize - ptr::from_ref(self.program.instruction(0).unwrap()) as usize;
+        offset_in_bytes / mem::size_of::<Instruction<T, W>>() as usize
     }
 
     // TODO: can overflow / underflow after an invalid instruction or free panic. Ordinarily, this will lead to VM termination (for an invalid instruction)
--- a/crates/vm2/src/decode.rs
+++ b/crates/vm2/src/decode.rs
@@ -18,34 +18,11 @@ use crate::{
     addressing_modes::{
         AbsoluteStack, AdvanceStackPointer, AnyDestination, AnySource, Arguments, CodePage,
         Immediate1, Immediate2, Register, Register1, Register2, RegisterAndImmediate,
-        RelativeStack, SourceWriter,
+        RelativeStack,
     },
-    instruction::{ExecutionEnd, ExecutionStatus},
     mode_requirements::ModeRequirements,
-    Instruction, Predicate, VirtualMachine, World,
+    Instruction, Predicate, World,
 };
-
-fn unimplemented_instruction<T, W>(variant: Opcode) -> Instruction<T, W> {
-    let mut arguments = Arguments::new(Predicate::Always, 0, ModeRequirements::none());
-    let variant_as_number: u16 = unsafe { std::mem::transmute(variant) };
-    Immediate1(variant_as_number).write_source(&mut arguments);
-    Instruction {
-        handler: unimplemented_handler,
-        arguments,
-    }
-}
-
-fn unimplemented_handler<T, W>(
-    vm: &mut VirtualMachine<T, W>,
-    _: &mut W,
-    _: &mut T,
-) -> ExecutionStatus {
-    let variant: Opcode = unsafe {
-        std::mem::transmute(Immediate1::get_u16(&(*vm.state.current_frame.pc).arguments))
-    };
-    eprintln!("Unimplemented instruction: {variant:?}");
-    ExecutionStatus::Stopped(ExecutionEnd::Panicked)
-}
 
 #[allow(clippy::too_many_lines)]
 pub(crate) fn decode<T: Tracer, W: World<T>>(raw: u64, is_bootloader: bool) -> Instruction<T, W> {
@@ -313,12 +290,22 @@ pub(crate) fn decode<T: Tracer, W: World<T>>(raw: u64, is_bootloader: bool) -> I
                     increment.then_some(out2),
                     arguments,
                 ),
-                zkevm_opcode_defs::UMAOpcode::StaticMemoryRead => unimplemented_instruction(
-                    Opcode::UMA(zkevm_opcode_defs::UMAOpcode::StaticMemoryRead),
-                ),
-                zkevm_opcode_defs::UMAOpcode::StaticMemoryWrite => unimplemented_instruction(
-                    Opcode::UMA(zkevm_opcode_defs::UMAOpcode::StaticMemoryWrite),
-                ),
+                zkevm_opcode_defs::UMAOpcode::StaticMemoryRead => {
+                    Instruction::from_static_memory_read(
+                        src1.try_into().unwrap(),
+                        out.try_into().unwrap(),
+                        increment.then_some(out2),
+                        arguments,
+                    )
+                }
+                zkevm_opcode_defs::UMAOpcode::StaticMemoryWrite => {
+                    Instruction::from_static_memory_write(
+                        src1.try_into().unwrap(),
+                        src2,
+                        increment.then_some(out.try_into().unwrap()),
+                        arguments,
+                    )
+                }
             }
         }
         Opcode::Invalid(_) => Instruction::from_invalid(),
--- a/crates/vm2/src/decommit.rs
+++ b/crates/vm2/src/decommit.rs
@@ -25,7 +25,8 @@ pub(crate) fn materialize_decommit_page<T: Tracer, W: World<T>>(
         return existing;
     }
 
-    let heap = vm.state.heaps.set_content_at(candidate_page, code);
+    vm.state.heaps.write_bytes(candidate_page, 0, code);
+    let heap = candidate_page;
     vm.world_diff.set_decommit_page(code_hash, heap);
 
     if heap != vm.state.current_frame.heap && heap != vm.state.current_frame.aux_heap {
--- a/crates/vm2/src/decommit.rs
+++ b/crates/vm2/src/decommit.rs
@@ -12,27 +12,31 @@ use crate::{
 
 /// Ensures that a decommit hash has a materialized reusable page and returns it.
 ///
-/// This page is pinned globally in [`WorldDiff`] and additionally kept alive in the
-/// bootloader frame (or current frame if no bootloader frame exists), matching
-/// decommit opcode teardown semantics.
+/// The resulting page is pinned globally in [`WorldDiff`]. If that page is not owned by the
+/// current frame, it is also recorded as kept alive in the bootloader frame (or current frame if
+/// no bootloader frame exists), matching decommit opcode teardown semantics.
 pub(crate) fn materialize_decommit_page<T: Tracer, W: World<T>>(
     vm: &mut VirtualMachine<T, W>,
     code_hash: U256,
     code: &[u8],
+    candidate_page: HeapId,
 ) -> HeapId {
     if let Some(existing) = vm.world_diff.decommit_page(code_hash) {
         return existing;
     }
 
-    let heap = vm.state.heaps.allocate_with_content(code);
+    let heap = vm.state.heaps.set_content_at(candidate_page, code);
     vm.world_diff.set_decommit_page(code_hash, heap);
 
-    let heaps_to_keep_alive = if let Some(bootloader_frame) = vm.state.previous_frames.first_mut() {
-        &mut bootloader_frame.heaps_i_am_keeping_alive
-    } else {
-        &mut vm.state.current_frame.heaps_i_am_keeping_alive
-    };
-    heaps_to_keep_alive.push(heap);
+    if heap != vm.state.current_frame.heap && heap != vm.state.current_frame.aux_heap {
+        let heaps_to_keep_alive =
+            if let Some(bootloader_frame) = vm.state.previous_frames.first_mut() {
+                &mut bootloader_frame.heaps_i_am_keeping_alive
+            } else {
+                &mut vm.state.current_frame.heaps_i_am_keeping_alive
+            };
+        heaps_to_keep_alive.push(heap);
+    }
 
     heap
 }
--- a/crates/vm2/src/decommit.rs
+++ b/crates/vm2/src/decommit.rs
@@ -4,7 +4,11 @@ use zkevm_opcode_defs::{
 };
 use zksync_vm2_interface::{CycleStats, Tracer};
 
-use crate::{program::Program, world_diff::WorldDiff, World};
+use crate::{
+    program::Program,
+    world_diff::{DecommitState, WorldDiff},
+    World,
+};
 
 impl WorldDiff {
     #[allow(clippy::too_many_arguments)]
@@ -77,7 +81,10 @@ impl WorldDiff {
         code_info[1] = 0;
         let code_key: U256 = U256::from_big_endian(&code_info);
 
-        let was_decommitted = self.decommitted_hashes.as_ref().get(&code_key) == Some(&true);
+        let was_decommitted = matches!(
+            self.decommitted_hashes.as_ref().get(&code_key),
+            Some(DecommitState::SucceededNoPage | DecommitState::SucceededWithPage(_))
+        );
         let cost = if was_decommitted {
             0
         } else {
@@ -97,7 +104,14 @@ impl WorldDiff {
         tracer: &mut T,
         code_hash: U256,
     ) -> (Vec<u8>, bool) {
-        let is_new = self.decommitted_hashes.insert(code_hash, true) != Some(true);
+        let is_new = match self.decommitted_hashes.as_ref().get(&code_hash).copied() {
+            None | Some(DecommitState::Unsuccessful) => {
+                self.decommitted_hashes
+                    .insert(code_hash, DecommitState::SucceededNoPage);
+                true
+            }
+            Some(DecommitState::SucceededNoPage | DecommitState::SucceededWithPage(_)) => false,
+        };
         let code = world.decommit_code(code_hash);
         if is_new {
             let code_len = u32::try_from(code.len()).expect("bytecode length overflow");
@@ -117,12 +131,30 @@ impl WorldDiff {
         if decommit.cost > *gas {
             // We intentionally record a decommitment event even if actual decommitment never happens because of an out-of-gas error.
             // This is how the old VM behaves.
-            self.decommitted_hashes.insert(decommit.code_key, false);
+            if !matches!(
+                self.decommitted_hashes.as_ref().get(&decommit.code_key),
+                Some(DecommitState::SucceededNoPage | DecommitState::SucceededWithPage(_))
+            ) {
+                self.decommitted_hashes
+                    .insert(decommit.code_key, DecommitState::Unsuccessful);
+            }
             // Unlike all other gas costs, this one is not paid if low on gas.
             return None;
         }
 
-        let is_new = self.decommitted_hashes.insert(decommit.code_key, true) != Some(true);
+        let is_new = match self
+            .decommitted_hashes
+            .as_ref()
+            .get(&decommit.code_key)
+            .copied()
+        {
+            None | Some(DecommitState::Unsuccessful) => {
+                self.decommitted_hashes
+                    .insert(decommit.code_key, DecommitState::SucceededNoPage);
+                true
+            }
+            Some(DecommitState::SucceededNoPage | DecommitState::SucceededWithPage(_)) => false,
+        };
         *gas -= decommit.cost;
 
         let decommit = world.decommit(decommit.code_key);
--- a/crates/vm2/src/decommit.rs
+++ b/crates/vm2/src/decommit.rs
@@ -2,13 +2,40 @@ use primitive_types::{H160, U256};
 use zkevm_opcode_defs::{
     ethereum_types::Address, system_params::DEPLOYER_SYSTEM_CONTRACT_ADDRESS_LOW,
 };
-use zksync_vm2_interface::{CycleStats, Tracer};
+use zksync_vm2_interface::{CycleStats, HeapId, Tracer};
 
 use crate::{
     program::Program,
     world_diff::{DecommitState, WorldDiff},
-    World,
+    VirtualMachine, World,
 };
+
+/// Ensures that a decommit hash has a materialized reusable page and returns it.
+///
+/// This page is pinned globally in [`WorldDiff`] and additionally kept alive in the
+/// bootloader frame (or current frame if no bootloader frame exists), matching
+/// decommit opcode teardown semantics.
+pub(crate) fn materialize_decommit_page<T: Tracer, W: World<T>>(
+    vm: &mut VirtualMachine<T, W>,
+    code_hash: U256,
+    code: &[u8],
+) -> HeapId {
+    if let Some(existing) = vm.world_diff.decommit_page(code_hash) {
+        return existing;
+    }
+
+    let heap = vm.state.heaps.allocate_with_content(code);
+    vm.world_diff.set_decommit_page(code_hash, heap);
+
+    let heaps_to_keep_alive = if let Some(bootloader_frame) = vm.state.previous_frames.first_mut() {
+        &mut bootloader_frame.heaps_i_am_keeping_alive
+    } else {
+        &mut vm.state.current_frame.heaps_i_am_keeping_alive
+    };
+    heaps_to_keep_alive.push(heap);
+
+    heap
+}
 
 impl WorldDiff {
     #[allow(clippy::too_many_arguments)]
@@ -81,10 +108,8 @@ impl WorldDiff {
         code_info[1] = 0;
         let code_key: U256 = U256::from_big_endian(&code_info);
 
-        let was_decommitted = matches!(
-            self.decommitted_hashes.as_ref().get(&code_key),
-            Some(DecommitState::SucceededNoPage | DecommitState::SucceededWithPage(_))
-        );
+        let decommit_state = self.decommitted_hashes.as_ref().get(&code_key).copied();
+        let was_decommitted = matches!(decommit_state, Some(DecommitState::Succeeded(_)));
         let cost = if was_decommitted {
             0
         } else {
@@ -92,7 +117,16 @@ impl WorldDiff {
             u32::from(code_length_in_words) * zkevm_opcode_defs::ERGS_PER_CODE_WORD_DECOMMITTMENT
         };
 
-        Some((UnpaidDecommit { cost, code_key }, is_evm))
+        let should_materialize = !was_decommitted;
+
+        Some((
+            UnpaidDecommit {
+                cost,
+                code_key,
+                should_materialize,
+            },
+            is_evm,
+        ))
     }
 
     /// Returns the decommitted contract code and a flag set to `true` if this is a fresh decommit (i.e.,
@@ -105,12 +139,8 @@ impl WorldDiff {
         code_hash: U256,
     ) -> (Vec<u8>, bool) {
         let is_new = match self.decommitted_hashes.as_ref().get(&code_hash).copied() {
-            None | Some(DecommitState::Unsuccessful) => {
-                self.decommitted_hashes
-                    .insert(code_hash, DecommitState::SucceededNoPage);
-                true
-            }
-            Some(DecommitState::SucceededNoPage | DecommitState::SucceededWithPage(_)) => false,
+            None | Some(DecommitState::Unsuccessful) => true,
+            Some(DecommitState::Succeeded(_)) => false,
         };
         let code = world.decommit_code(code_hash);
         if is_new {
@@ -133,7 +163,7 @@ impl WorldDiff {
             // This is how the old VM behaves.
             if !matches!(
                 self.decommitted_hashes.as_ref().get(&decommit.code_key),
-                Some(DecommitState::SucceededNoPage | DecommitState::SucceededWithPage(_))
+                Some(DecommitState::Succeeded(_))
             ) {
                 self.decommitted_hashes
                     .insert(decommit.code_key, DecommitState::Unsuccessful);
@@ -148,12 +178,8 @@ impl WorldDiff {
             .get(&decommit.code_key)
             .copied()
         {
-            None | Some(DecommitState::Unsuccessful) => {
-                self.decommitted_hashes
-                    .insert(decommit.code_key, DecommitState::SucceededNoPage);
-                true
-            }
-            Some(DecommitState::SucceededNoPage | DecommitState::SucceededWithPage(_)) => false,
+            None | Some(DecommitState::Unsuccessful) => true,
+            Some(DecommitState::Succeeded(_)) => false,
         };
         *gas -= decommit.cost;
 
@@ -173,6 +199,18 @@ impl WorldDiff {
 pub(crate) struct UnpaidDecommit {
     cost: u32,
     code_key: U256,
+    /// Indicates whether the decommit page should be materialized after successful payment.
+    should_materialize: bool,
+}
+
+impl UnpaidDecommit {
+    pub(crate) const fn code_key(self) -> U256 {
+        self.code_key
+    }
+
+    pub(crate) const fn should_materialize(self) -> bool {
+        self.should_materialize
+    }
 }
 
 pub(crate) fn u256_into_address(source: U256) -> H160 {
--- a/crates/vm2/src/decommit.rs
+++ b/crates/vm2/src/decommit.rs
@@ -7,6 +7,7 @@ use zksync_vm2_interface::{CycleStats, Tracer};
 use crate::{program::Program, world_diff::WorldDiff, World};
 
 impl WorldDiff {
+    #[allow(clippy::too_many_arguments)]
     pub(crate) fn decommit<T: Tracer>(
         &mut self,
         world: &mut impl World<T>,
--- a/crates/vm2/src/decommit.rs
+++ b/crates/vm2/src/decommit.rs
@@ -15,6 +15,7 @@ impl WorldDiff {
         default_aa_code_hash: [u8; 32],
         evm_interpreter_code_hash: [u8; 32],
         is_constructor_call: bool,
+        tx_number_in_block: u16,
     ) -> Option<(UnpaidDecommit, bool)> {
         let deployer_system_contract_address =
             Address::from_low_u64_be(DEPLOYER_SYSTEM_CONTRACT_ADDRESS_LOW.into());
@@ -27,6 +28,7 @@ impl WorldDiff {
                 tracer,
                 deployer_system_contract_address,
                 address,
+                tx_number_in_block,
             );
             let mut code_info_bytes = [0; 32];
             code_info.to_big_endian(&mut code_info_bytes);
--- a/crates/vm2/src/decommit.rs
+++ b/crates/vm2/src/decommit.rs
@@ -159,8 +159,9 @@ impl WorldDiff {
         gas: &mut u32,
     ) -> Option<Program<T, W>> {
         if decommit.cost > *gas {
-            // We intentionally record a decommitment event even if actual decommitment never happens because of an out-of-gas error.
-            // This is how the old VM behaves.
+            // We intentionally keep this hash visible even though decommit did not execute.
+            // Legacy VM includes far-call OOG attempts in "used contracts", and shadow mode
+            // compares that output.
             if !matches!(
                 self.decommitted_hashes.as_ref().get(&decommit.code_key),
                 Some(DecommitState::Succeeded(_))
--- a/crates/vm2/src/heap.rs
+++ b/crates/vm2/src/heap.rs
@@ -1,5 +1,5 @@
 use std::{
-    fmt, mem,
+    fmt,
     ops::{Index, Range},
 };
 
@@ -78,11 +78,6 @@ impl Heap {
         for page in self.pages.into_iter().flatten() {
             pagepool.recycle_page(page);
         }
-    }
-
-    fn replace_contents(&mut self, bytes: &[u8], pagepool: &mut PagePool) {
-        let old = mem::replace(self, Self::from_bytes(bytes, pagepool));
-        old.recycle(pagepool);
     }
 
     pub(crate) fn read_u256(&self, start_address: u32) -> U256 {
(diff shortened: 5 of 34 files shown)


crates/vm2/src/instruction_handlers/common.rs
```diff
@@ -46,6 +47,16 @@ pub(crate) fn full_boilerplate<Opcode: OpcodeType, T: Tracer, W: World<T>>(
 ) -> ExecutionStatus {
     let args = unsafe { &(*vm.state.current_frame.pc).arguments };
 
+    // We are not checking whether the callstack is full here. However,
```
popzxc, line 50: In theory, we should panic on call stack saturation per zk_evm, but in practice the callstack is so big that it doesn't make sense to spend cycles on this check IMHO.

- zk_evm reference:
  - [crates/zk_evm/src/vm_state/cycle.rs#L199-L223](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/vm_state/cycle.rs#L199-L223)

crates/vm2/src/instruction_handlers/decommit.rs
```diff
@@ -38,8 +38,34 @@ fn decommit<T: Tracer, W: World<T>>(
             vm.state.current_frame.gas += extra_cost;
         }
 
-        let heap = vm.state.heaps.allocate_with_content(program.as_ref());
-        vm.state.current_frame.heaps_i_am_keeping_alive.push(heap);
+        let heap = if is_fresh {
```
popzxc: We used to always allocate a new page, but `zk_evm` marks decommitted pages as persisted and not-to-be-freed. At the same time, we account for the fact that one could pay for decommitment without assigning a decommit page.

- zk_evm reference:
  - [crates/zk_evm/src/reference_impls/decommitter.rs#L61-L69](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/reference_impls/decommitter.rs#L61-L69)
  - [crates/zk_evm/src/reference_impls/memory.rs#L392](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/reference_impls/memory.rs#L392)

crates/vm2/src/instruction_handlers/far_call.rs
```diff
@@ -60,20 +60,29 @@ where
...
-                abi.is_constructor_call,
-            );
-
-            // calldata has to be constructed even if we already know we will panic because
-            // overflowing start + length makes the heap resize even when already panicking.
-            let already_failed = decommit_result.is_none() || IS_SHARD && abi.shard_id != 0;
+            let shard_call_failed = IS_SHARD && abi.shard_id != 0;
 
-            let maybe_calldata = get_calldata(raw_abi, raw_abi_is_pointer, vm, already_failed);
+            let (maybe_calldata, decommit_result) = if shard_call_failed {
```
popzxc, line 67: If a non-zero shard is called, we should fail the call without doing decommitment.

- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/far_call.rs#L110-L164](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/far_call.rs#L110-L164)

crates/vm2/src/instruction_handlers/heap_access.rs
```diff
@@ -17,6 +17,12 @@ use crate::{
     ExecutionEnd, Instruction, VirtualMachine, World,
 };
 
+/// Dedicated heap slot used to model `zk_evm` static memory.
```
popzxc: Support for static memory read/write.
 
- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113)
  - [crates/zk_evm/src/opcodes/execution/uma.rs#L109-L113](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/uma.rs#L109-L113)
  - [crates/zk_evm/src/opcodes/execution/log.rs#L315-L335](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L315-L335)

crates/vm2/src/instruction_handlers/heap_access.rs
```diff
@@ -190,6 +196,62 @@ fn load_pointer<T: Tracer, W: World<T>, const INCREMENT: bool>(
     })
 }
 
+fn load_static<T: Tracer, W: World<T>, In: Source, const INCREMENT: bool>(
```
popzxc, line 197: Support for static memory read/write.
 
- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113)
  - [crates/zk_evm/src/opcodes/execution/uma.rs#L109-L113](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/uma.rs#L109-L113)
  - [crates/zk_evm/src/opcodes/execution/log.rs#L315-L335](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L315-L335)

crates/vm2/src/instruction_handlers/heap_access.rs
```diff
@@ -211,6 +273,26 @@ impl<T: Tracer, W: World<T>> Instruction<T, W> {
         Self::from_read::<AuxHeap>(src, out, incremented_out, arguments)
     }
 
+    /// Creates a static memory read instruction.
```
popzxc, line 274: Support for static memory read/write.
 
- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113)
  - [crates/zk_evm/src/opcodes/execution/uma.rs#L109-L113](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/uma.rs#L109-L113)
  - [crates/zk_evm/src/opcodes/execution/log.rs#L315-L335](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L315-L335)

crates/vm2/src/instruction_handlers/heap_access.rs
```diff
@@ -251,6 +333,23 @@ impl<T: Tracer, W: World<T>> Instruction<T, W> {
         Self::from_write::<AuxHeap>(src1, src2, incremented_out, arguments, false)
     }
 
+    /// Creates a static memory write instruction.
```
popzxc, line 334: Support for static memory read/write.
 
- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113)
  - [crates/zk_evm/src/opcodes/execution/uma.rs#L109-L113](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/uma.rs#L109-L113)
  - [crates/zk_evm/src/opcodes/execution/log.rs#L315-L335](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L315-L335)

crates/vm2/src/instruction_handlers/precompiles.rs
```diff
@@ -78,7 +78,7 @@ fn precompile_call<T: Tracer, W: World<T>>(
             // This is safe because system contracts are trusted
             let aux_data = PrecompileAuxData::from_u256(Register2::get(args, &mut vm.state));
             let Ok(()) = vm.state.use_gas(aux_data.extra_ergs_cost) else {
-                vm.state.current_frame.pc = spontaneous_panic();
```
popzxc, line 81: On insufficient extra ergs we should not panic.

- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/log.rs#L88-L134](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L88-L134)
  - [crates/zk_evm/src/opcodes/execution/log.rs#L294-L305](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L294-L305)

crates/vm2/src/instruction_handlers/ret.rs
```diff
@@ -45,8 +45,15 @@ fn naked_ret<T: Tracer, W: World<T>, RT: TypeLevelReturnType, const TO_LABEL: bo
         } else {
             let (raw_abi, is_pointer) = Register1::get_with_pointer_flag(args, &mut vm.state);
             let result = get_calldata(raw_abi, is_pointer, vm, false).filter(|pointer| {
-                vm.state.current_frame.is_kernel
-                    || pointer.memory_page != vm.state.current_frame.calldata_heap
+                if vm.state.current_frame.is_kernel {
```
popzxc, line 49: - zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/ret.rs#L58-L75](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/ret.rs#L58-L75)

crates/vm2/src/instruction_handlers/storage.rs
```diff
@@ -18,9 +18,14 @@ fn sstore<T: Tracer, W: World<T>>(
         let key = Register1::get(args, &mut vm.state);
         let value = Register2::get(args, &mut vm.state);
 
-        let refund =
-            vm.world_diff
-                .write_storage(world, tracer, vm.state.current_frame.address, key, value);
+        let refund = vm.world_diff.write_storage(
```
popzxc, line 21: Append tx number to be included in storage logs.

crates/vm2/src/instruction_handlers/storage.rs
```diff
@@ -49,9 +54,13 @@ fn sload<T: Tracer, W: World<T>>(
     boilerplate_ext::<opcodes::StorageRead, _, _>(vm, world, tracer, |vm, args, world, tracer| {
         let key = Register1::get(args, &mut vm.state);
 
-        let (value, refund) =
-            vm.world_diff
-                .read_storage(world, tracer, vm.state.current_frame.address, key);
+        let (value, refund) = vm.world_diff.read_storage(
```
popzxc, line 57: Append tx number to be included in storage logs.

crates/vm2/src/tests/divergence_regressions.rs
popzxc, line 1: Tests here were mainly added to test behavior vs zk_evm and check regressions. Basically, the flow was:
1. Find a divergence
2. Add a test to trigger it
3. Create a fix in the codebase.

crates/vm2/src/callframe.rs
```diff
@@ -141,17 +141,20 @@ impl<T: Tracer, W: World<T>> Callframe<T, W> {
 
     /// Gets a raw inferred program counter. This value can be garbage if the frame is on an invalid instruction or free panic.
     #[allow(clippy::cast_possible_wrap)] // false positive: `Instruction` isn't that large
-    pub(crate) fn get_raw_pc(&self) -> isize {
+    pub(crate) fn get_raw_pc(&self) -> usize {
```
popzxc, line 144: This one is not a divergence, but requirement for RISC-V compilation -- we don't have signed division implemented.

crates/vm2/src/decode.rs
```diff
@@ -313,12 +290,22 @@ pub(crate) fn decode<T: Tracer, W: World<T>>(raw: u64, is_bootloader: bool) -> I
                     increment.then_some(out2),
                     arguments,
                 ),
-                zkevm_opcode_defs::UMAOpcode::StaticMemoryRead => unimplemented_instruction(
-                    Opcode::UMA(zkevm_opcode_defs::UMAOpcode::StaticMemoryRead),
-                ),
-                zkevm_opcode_defs::UMAOpcode::StaticMemoryWrite => unimplemented_instruction(
-                    Opcode::UMA(zkevm_opcode_defs::UMAOpcode::StaticMemoryWrite),
-                ),
+                zkevm_opcode_defs::UMAOpcode::StaticMemoryRead => {
```
popzxc, line 293: - zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/uma.rs#L81-L113)

crates/vm2/src/decommit.rs
```diff
@@ -15,6 +16,7 @@ impl WorldDiff {
         default_aa_code_hash: [u8; 32],
         evm_interpreter_code_hash: [u8; 32],
         is_constructor_call: bool,
+        tx_number_in_block: u16,
```
popzxc, line 55: For storage logs

crates/vm2/src/state.rs
```diff
@@ -119,7 +119,11 @@ impl<T: Tracer, W: World<T>> State<T, W> {
         }
     }
 
-    pub(crate) fn rollback(&mut self, snapshot: StateSnapshot) {
+    pub(crate) fn rollback(
+        &mut self,
+        snapshot: StateSnapshot,
+        mut is_heap_pinned: impl FnMut(HeapId) -> bool,
```
popzxc, line 143: To preserve decommitted pages; I don't really like this solution, but I wasn't able to come up with an elegant solution that would not require a big refactoring, and this PR is already has too much going on, so I went with the simple option.

crates/vm2/src/tracing.rs
```diff
@@ -284,20 +284,14 @@ impl<T: Tracer, W: World<T>> CallframeInterface for CallframeWrapper<'_, T, W> {
...
+    #[allow(clippy::cast_possible_truncation)]
     fn program_counter(&self) -> Option<u16> {
         if let Some(call) = self.near_call_on_top() {
             Some(call.previous_frame_pc)
         } else {
             let offset = self.frame.get_raw_pc();
-            if offset < 0
-                || offset > u16::MAX as isize
-                || self.frame.program.instruction(offset as u16).is_none()
+            if offset > u16::MAX as usize || self.frame.program.instruction(offset as u16).is_none()
```
popzxc, line 294: In the situations where we used to have `offset < 0` we now should have `offset > u16::MAX` (because we use `usize::MAX` in `get_raw_pc()` as a fallback), so I expect no changes in behavior.

crates/vm2/src/vm.rs
```diff
@@ -161,7 +161,9 @@ impl<T: Tracer, W: World<T>> VirtualMachine<T, W> {
             .take()
             .expect("`rollback()` called without a snapshot");
         self.world_diff.external_rollback(snapshot.world_snapshot);
-        self.state.rollback(snapshot.state_snapshot);
+        self.state.rollback(snapshot.state_snapshot, |heap| {
```
popzxc, line 165: Preserving pinned decommit pages.

crates/vm2/src/vm.rs
```diff
@@ -238,38 +240,38 @@ impl<T: Tracer, W> VirtualMachine<T, W> {
...
-                self.state.current_frame.heap,
-                self.state.current_frame.aux_heap,
-            ]
-            .iter()
-            .chain(&self.state.current_frame.heaps_i_am_keeping_alive)
-            {
-                if Some(heap) != heap_to_keep {
-                    self.state.heaps.deallocate(heap);
-                }
+        let mut frame = self.state.previous_frames.pop()?;
```
popzxc, line 250: Same -- preserving pinned decommit pages.

crates/vm2/src/world_diff.rs
```diff
@@ -25,12 +26,23 @@ pub struct WorldDiff {
     pub(crate) pubdata: RollbackablePod<i32>,
     storage_refunds: RollbackableLog<u32>,
     pubdata_costs: RollbackableLog<i32>,
+    storage_logs: Vec<LogQuery>,
+    rollback_storage_logs: Vec<LogQuery>,
```
popzxc, line 30: Storage access logs & rollback markers -- needed to actually verify the execution in `tee_verifier`

- zk_evm reference:
  - [crates/zk_evm/src/opcodes/execution/log.rs#L143-L178](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L143-L178)
  - [crates/zk_evm/src/opcodes/execution/log.rs#L199-L236](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L199-L236)
  - [crates/zk_evm/src/vm_state/helpers.rs#L148-L176](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/vm_state/helpers.rs#L148-L176)

crates/vm2/src/world_diff.rs
```diff
@@ -180,6 +236,21 @@ impl WorldDiff {
         self.pubdata_costs.as_ref()
     }
 
+    /// Returns all recorded storage log queries.
+    ///
+    /// These logs are sufficient for vm2 state-transition checks and diagnostics.
+    // TODO: We don't fill all the `zk_evm` witness metadata, so this is not suitable for
```
popzxc, line 261: What's implemented is bare minimum required to verify execution, we still cannot generate witness with this data -- but that is not a goal anyway.

Cargo.toml
```diff
@@ -30,9 +30,9 @@ primitive-types = "0.12.1"
 proptest = "1.4"
 
 # "Internal" dependencies
-zkevm_opcode_defs = "=0.153.9"
-zk_evm_abstractions = "=0.153.9"
-zk_evm = "=0.153.9"
+zkevm_opcode_defs = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
+zk_evm_abstractions = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
+zk_evm = { git = "https://github.com/matter-labs/zksync-protocol", branch = "popzxc-airbender-precompiles" } # TODO: Must be pinned before merge
```
popzxc, line 35: This is to be replaced before merge, but as a heads up -- we'll likely have to stick with git deps now since we now have a dependency on airbender. This is to be discussed internally (I'll start a thread).

crates/vm2/src/world_diff.rs
slowli, line 30: This looks unfortunate, IMO; optimally, it should be world responsibility to keep these logs if desired (which I've discussed with Joonathan previously 🙃). Pragmatically, it's unrealistic to refactor the `World` responsibilities rn.

crates/vm2/src/instruction_handlers/decommit.rs
```diff
@@ -38,8 +38,34 @@ fn decommit<T: Tracer, W: World<T>>(
...
+
+        // Decommit page mapping lives for the whole VM run, so nested-frame decommits
+        // must pin pages in the bootloader frame rather than in the current frame.
+        let heaps_to_keep_alive =
+            if let Some(bootloader_frame) = vm.state.previous_frames.first_mut() {
+                &mut bootloader_frame.heaps_i_am_keeping_alive
+            } else {
+                &mut vm.state.current_frame.heaps_i_am_keeping_alive
+            };
+        if !heaps_to_keep_alive.contains(&heap) {
```
slowli: Nit: This check looks like a potential performance regression cause. Intuitively, `heaps_to_keep_alive.contains(&heap)` *should* be determined by `let heap = _` logic above (e.g., if `is_fresh`, then the heap is newly allocated, so it shouldn't be present in `heaps_i_am_keeping_alive`.

crates/vm2/src/world_diff.rs
```diff
@@ -25,12 +26,23 @@ pub struct WorldDiff {
...
 
     // The fields below are only rolled back when the whole VM is rolled back.
     /// Values indicate whether a bytecode was successfully decommitted. When accessing decommitted hashes
     /// for the execution state, we need to track both successful and failed decommitments; OTOH, only successful ones
     /// matter when computing decommitment cost.
     pub(crate) decommitted_hashes: RollbackableMap<U256, bool>,
+    /// Memory page assigned to each successfully decommitted hash by the `Decommit` opcode.
+    ///
+    /// Like `decommitted_hashes`, this is rolled back only by external VM snapshots.
+    decommit_pages: RollbackableMap<U256, u32>,
```
slowli: I wonder whether this can be united with `decommitted_hashes` into a single map (`RollbackableMap<U256, Option<u32>>`), with values corresponding to heap IDs of successful decommitments.

crates/vm2/src/instruction_handlers/heap_access.rs
```diff
@@ -190,6 +196,62 @@ fn load_pointer<T: Tracer, W: World<T>, const INCREMENT: bool>(
     })
 }
 
+fn load_static<T: Tracer, W: World<T>, In: Source, const INCREMENT: bool>(
+    vm: &mut VirtualMachine<T, W>,
+    world: &mut W,
+    tracer: &mut T,
+) -> ExecutionStatus {
+    boilerplate::<opcodes::HeapRead, _, _>(vm, world, tracer, |vm, args| {
```
slowli: Shouldn't this use a new opcode instead of `HeapRead`? Ditto for the `store_static` handler below.

slowli (commented):
Couple of quick comments; will try to do a more thorough review shortly.

crates/vm2/src/instruction_handlers/decommit.rs
popzxc: Implemented in [a351fbd](https://github.com/matter-labs/vm2/pull/95/commits/a351fbd8218634a720dcc0f7a16d32b4586e1031)

crates/vm2/src/instruction_handlers/heap_access.rs
popzxc: Implemented in [2e72cc1](https://github.com/matter-labs/vm2/pull/95/commits/2e72cc14ea4cd6282cacd9a954b66fdb7aa7f236)

crates/vm2/src/world_diff.rs
popzxc: Good point. Though we have three states: failed decommit, decommit with no page, and decommit with page, so I added an enum instead. Implemented in [e4a4fb4](https://github.com/matter-labs/vm2/pull/95/commits/e4a4fb44b43f77ec4fba3127dfb02fb20c691ece)

crates/vm2/src/instruction_handlers/decommit.rs
```diff
@@ -38,8 +38,35 @@ fn decommit<T: Tracer, W: World<T>>(
             vm.state.current_frame.gas += extra_cost;
         }
 
-        let heap = vm.state.heaps.allocate_with_content(program.as_ref());
-        vm.state.current_frame.heaps_i_am_keeping_alive.push(heap);
+        // We only need to mutate keep-alive state when this opcode allocates a new heap.
+        let (heap, should_record_keep_alive) = if is_fresh {
+            let heap = vm.state.heaps.allocate_with_content(program.as_ref());
```
StanislavBreadless: how does it work? (maybe I miss something)

`allocate_with_content`:

```
    pub(crate) fn allocate_with_content(&mut self, content: &[u8]) -> HeapId {
        let id = self.allocate();
        self.read
            .get_mut(id)
            .write_u256(0, U256::from_big_endian(content));
        id
    }
```

`U256::from_big_endian(content)` so we assume that the content is U256, but in reality it is a big slice of bytecode

crates/vm2/src/instruction_handlers/decommit.rs
```diff
@@ -38,8 +38,35 @@ fn decommit<T: Tracer, W: World<T>>(
...
+            if let Some(heap) = vm.world_diff.decommit_page(code_hash) {
+                (heap, false)
+            } else {
+                let heap = vm.state.heaps.allocate_with_content(program.as_ref());
+                vm.world_diff.set_decommit_page(code_hash, heap);
+                (heap, true)
+            }
+        };
+
+        // Decommit page mapping lives for the whole VM run, so nested-frame decommits
```
StanislavBreadless: We have the following logic inside pop_frame:

```
        for &heap in [
            self.state.current_frame.heap,
            self.state.current_frame.aux_heap,
        ]
        .iter()
        .chain(&self.state.current_frame.heaps_i_am_keeping_alive)
        {
            if Some(heap) != heap_to_keep && !self.world_diff.is_decommit_page_pinned(heap) {
                self.state.heaps.deallocate(heap);
            }
        }
```

Do I understand it correctly that since you've added the check here, we can remove the `!self.world_diff.is_decommit_page_pinned(heap)` part? Since decommitted pages are never part of `self.state.current_frame.heaps_i_am_keeping_alive` unless `state.current_frame` is the bootloader at which point we dont care?

crates/vm2/src/instruction_handlers/decommit.rs
```diff
@@ -38,8 +38,35 @@ fn decommit<T: Tracer, W: World<T>>(
...
+            (heap, true)
+        } else {
+            // `pay_for_decommit` can mark the hash as decommitted without assigning a memory page.
+            // In that case, the first `Decommit` opcode materializes the page lazily and starts
+            // tracking it for frame teardown semantics.
+            if let Some(heap) = vm.world_diff.decommit_page(code_hash) {
+                (heap, false)
+            } else {
+                let heap = vm.state.heaps.allocate_with_content(program.as_ref());
+                vm.world_diff.set_decommit_page(code_hash, heap);
```
StanislavBreadless: are you sure this is aligned with the previous implementation? IIUC in old vm, if the following sequence sequence happened:

- far call calls a contract (decommitted into the "code" area)
- someone tries to decommit the code

the pointer to the "code" area has been returned. 

Right now, we materialize it into the heap, so we get two copies of the same code

StanislavBreadless: What's more, it means we get two pages where the code is decomitted. It is not *too* bad, but the user paid for only one decommit, which is not nice

crates/vm2/src/instruction_handlers/far_call.rs
```diff
@@ -60,20 +60,29 @@ where
...
-                destination_address,
-                vm.settings.default_aa_code_hash,
-                vm.settings.evm_interpreter_code_hash,
-                abi.is_constructor_call,
-            );
-
-            // calldata has to be constructed even if we already know we will panic because
-            // overflowing start + length makes the heap resize even when already panicking.
-            let already_failed = decommit_result.is_none() || IS_SHARD && abi.shard_id != 0;
+            let shard_call_failed = IS_SHARD && abi.shard_id != 0;
```
StanislavBreadless, line 65: nit: maybe just always validate `abi.shard_id != 0`? we dont support any shard except for zero anyway

though I dont have enough knowledge of the codebase to judge whether this optimization wont break anything

crates/vm2/src/instruction_handlers/storage.rs
StanislavBreadless: @popzxc why didnt we need it before?

crates/vm2/src/callframe.rs
```diff
@@ -141,17 +141,20 @@ impl<T: Tracer, W: World<T>> Callframe<T, W> {
...
         // We cannot use `<*const _>::offset_from` because `self.pc` isn't guaranteed to be allocated within `self.program`
         // (invalid instructions and free panics aren't).
-        let offset_in_bytes =
-            self.pc as isize - ptr::from_ref(self.program.instruction(0).unwrap()) as isize;
-        offset_in_bytes / mem::size_of::<Instruction<T, W>>() as isize
+        // We cannot use `isize` either, because it causes troubles when compiled to `RISC-V`, so we are
+        // returning the value that is clearly out of scope.
```
StanislavBreadless, line 148: nit: can you maybe add some const assert to ensure that it is out of bounds?

crates/vm2/src/world_diff.rs
```diff
@@ -25,12 +26,24 @@ pub struct WorldDiff {
...
+    rollback_storage_logs: Vec<LogQuery>,
 
     // The fields below are only rolled back when the whole VM is rolled back.
-    /// Values indicate whether a bytecode was successfully decommitted. When accessing decommitted hashes
-    /// for the execution state, we need to track both successful and failed decommitments; OTOH, only successful ones
-    /// matter when computing decommitment cost.
-    pub(crate) decommitted_hashes: RollbackableMap<U256, bool>,
+    /// Tracks decommit attempts and successful decommit state for each bytecode hash.
+    ///
+    /// This includes failed attempts (out of gas) because they affect old-VM-compatible behavior,
```
StanislavBreadless: It is either wrong or maybe the comment has to be clarified:

- if a *decommit* has failed (i.e. user did not have enough gas for the decommit itself), then the decommit should never be tracked and totally fail, since otherwise users could spam bad decommits 
- however, if decommit has been successful and the frame with it has reverted, then yes we need to remember it

crates/vm2/src/world_diff.rs
```diff
@@ -25,12 +26,24 @@ pub struct WorldDiff {
...
+    /// and successful attempts may or may not already have a materialized decommit page.
```
StanislavBreadless: as mentioned in the comments above, 99% chance that it is not aligned with the previous behavior even though it may not be too bad

crates/vm2/src/world_diff.rs
```diff
@@ -86,14 +124,29 @@ impl WorldDiff {
...
         let newly_added = self.read_storage_slots.add((contract, key));
         if newly_added {
             tracer.on_extra_prover_cycles(CycleStats::StorageRead);
         }
 
         self.pubdata_costs.push(0);
-        (self.just_read_storage(world, contract, key), newly_added)
+        let value = self.just_read_storage(world, contract, key);
+        self.storage_logs.push(LogQuery {
+            timestamp: Timestamp(u32::try_from(self.storage_logs.len()).unwrap_or(u32::MAX)),
```
StanislavBreadless: I would personally panic with `unwrap()` here tbh since it is totally unexpected IIUC

StanislavBreadless: the behavior for timestamp here is different from how it was done in the old vm, where timestamp was assigned per cycle:
```
    pub fn increment_timestamp_after_cycle(&mut self) {
        self.local_state.timestamp += zkevm_opcode_defs::TIME_DELTA_PER_CYCLE
    }
```

I dont think it matters all that much as long as there is only one log (and maybe its reverted one) per timestamp.

Judging from the code storage_logs never get rolled back, but maybe it makes sense to add a comment about that it is crucial for the uniqueness of timestamps

crates/vm2/src/world_diff.rs
```diff
@@ -288,11 +400,21 @@ impl WorldDiff {
             l2_to_l1_logs: self.l2_to_l1_logs.snapshot(),
             transient_storage_changes: self.transient_storage_changes.snapshot(),
             pubdata: self.pubdata.snapshot(),
+            storage_logs_len: self.storage_logs.len(),
+            rollback_storage_logs_len: self.rollback_storage_logs.len(),
         }
     }
 
     #[allow(clippy::needless_pass_by_value)] // intentional: we require a snapshot to be rolled back to no more than once
     pub(crate) fn rollback(&mut self, snapshot: Snapshot) {
```
StanislavBreadless: this function is used in two places:

```
  1. world_diff.rs:451 — from external_rollback, which is itself called from VirtualMachine::rollback at vm.rs:163
  2. ret.rs:117 — directly on far-call return failure (revert/panic), to undo the child frame's state changes
```

The new behavior (appending `rollback_storage_logs` to `storage_logs`) is okay only for the second use-case.

For the first one, everything *does* turn out to be correct, but it looks like a missed shot into leg: we are fine because this `rollback` is called before `   self.storage_logs.truncate(snapshot.storage_logs_len);`. 

I would personally separate the two functions to avoid any potential mistakes. 

PS: we now have `storage_logs_len`/`rollback_storage_logs_len` in both external and internal snapshot, maybe makes sense to only use the internal one

crates/vm2/src/instruction_handlers/decommit.rs
popzxc: vm2 has a tricky feature `single_instruction_test` that [overrides paths for some modules](https://github.com/matter-labs/vm2/blob/master/crates/vm2/src/lib.rs#L11), and it looks like it confuses rust-analyzer a lot -- even though this feature is not enabled by default, `rust-analyzer` picks it up and it messes up all the "goto" commands.

I assume that you are referring to the implementation in [vm2/src/single_instruction_test/heap.rs](https://github.com/matter-labs/vm2/blob/master/crates/vm2/src/single_instruction_test/heap.rs#L72), which is an implementation for a test.
The actual implementation is in [vm2/src/heap.rs](https://github.com/matter-labs/vm2/blob/master/crates/vm2/src/heap.rs#L220), which properly creates a heap from bytes.

crates/vm2/src/instruction_handlers/storage.rs
popzxc: We don't really need it -- I just thought that it might be useful in case we'll need to do some debugging. If some internals of VM2 are behaving incorrectly, the ability to tell which transaction caused the suspicious log can be helpful.

Given that this change is pretty trivial, I thought that it's fine to add.

crates/vm2/src/world_diff.rs
popzxc: OK, this one is funny. This behavior is both correct and present in `master` branch, and the wrong party here is old vm (or rather, decommitter implementation for it in multivm).
There is even a [test for that specific behavior](https://github.com/matter-labs/vm2/blob/master/crates/vm2/src/tests/far_call_decommitment.rs#L152) in the master branch.

Here's the story.
In both fast and old VM we expose `CurrentExecutionState` as a part of the output state: [FastVM](https://github.com/matter-labs/zksync-era/blob/main/core/lib/multivm/src/versions/vm_fast/vm.rs#L354) / [LegacyVM](https://github.com/matter-labs/zksync-era/blob/main/core/lib/multivm/src/versions/vm_latest/vm.rs#L123). There, we expose `used_contract_hashes`.

We need this to know which preimages we need to pass to the witness.

Now, the decommitter for the legacy VM has a quirk: it [stores](https://github.com/matter-labs/zksync-era/blob/main/core/lib/multivm/src/versions/vm_latest/old_vm/oracles/decommitter.rs#L207) `None` in decommitted code hashes as a marker before you pay for decommitment -- because we call `prepare_for_decommitment` before we pay for it. Note that this only affects far calls, since for Decommit we only call `prepare_to_decommit` [*after* we've checked for gas](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L397). 

And in order to get used contracts, [we simply iterate through the keys, which include `None` values](https://github.com/matter-labs/zksync-era/blob/main/core/lib/multivm/src/versions/vm_latest/implementation/statistics.rs#L44).

Finally, in shadow vm we [check divergencies](https://github.com/matter-labs/zksync-era/blob/main/core/lib/vm_interface/src/utils/shadow.rs#L111-L150) for CurrentExecutionState.

So I assume what happened in the past was:
- There was no this logic there initially.
- When a batch with OOG on decommit occurred, ShadowVM caught a divergence.
- It was implemented in vm2 to match the behavior.

In theory, a better fix would've been to fix a bug in the old decommitter, since as far as I can tell it shouldn't affect anything, but I guess changing protocol-related behavior without an upgrade is scary, so mimicking this in vm2 was probably a right call :) 

And it looks like my research matches with the PRs in this repo: [PR to record an event on OOG decommitment](https://github.com/matter-labs/vm2/pull/56), [Follow-up PR](https://github.com/matter-labs/vm2/pull/57).

cc @slowli in case I got something wrong.

popzxc: I will clarify the comment though.

crates/vm2/src/instruction_handlers/decommit.rs
popzxc: This is a good catch. I don't think that we can call it a divergence -- this is an artifact of _fixing_ a divergence, which, to the best of my knowledge, should not have any observable implications and is not that bad given that decommitment in airbender is cheap. But it's certainly an inefficiency -- will fix that.

And actually not having to lazily materialize it probably will let me to simplify the code somewhat.

crates/vm2/src/instruction_handlers/far_call.rs
popzxc: I think it's better to keep it in a way that is somewhat aligned with the reference impl; I don't think that this extra check will have any noticeable impact on the cycle count.

crates/vm2/src/world_diff.rs
popzxc: Yup, I'm changing the code to eagerly materialize the page on decommit, so we won't have this enum variant anymore.

crates/vm2/src/world_diff.rs
popzxc: Yup, I've done it this way to not mimic more logic that is not needed (as we're not generating witness) and it only enforces uniqueness. I'll add a comment.

crates/vm2/src/world_diff.rs
popzxc: That's a good catch! Will fix.

crates/vm2/src/instruction_handlers/decommit.rs
StanislavBreadless: It seems that:
- whenever we create a new frame, we allocate heap/auxheap
- but also when `decommit` is called a NEW heap is created with just the decommitted data.

This is fine, but it is a divergence from the past (where we would've reused the actual heap of the contract). just a FYI

crates/vm2/src/world_diff.rs
StanislavBreadless: Got it, thanks!

crates/vm2/src/callframe.rs
popzxc: Not really, since `Instruction` is generic. But since it needs to be outside of `u16` range, on a 32-bit platform, the size of the instruction should be 65537 bytes or more for this code not be out of range (2^32-1 / 2^16 -1).

crates/vm2/src/instruction_handlers/decommit.rs
popzxc: I think it's better to keep both checks in place. In kernel space, we can return far pointer to older heaps, which would put the heap to `heaps_i_am_keeping_alive` for the parent frame, and which, in its turn, would make this heap a candidate for deallocation. The likelihood of that happening is negligible (unless we do such trickery somewhere in system contracts, which I highly doubt), but technically this check is needed for correctness.

crates/vm2/src/instruction_handlers/decommit.rs
popzxc: Fixed in [8af6473](https://github.com/matter-labs/vm2/pull/95/commits/8af6473bc8c0f967e60353b49388808c5cacb66e)

crates/vm2/src/world_diff.rs
popzxc: Fixed in [8af6473](https://github.com/matter-labs/vm2/pull/95/commits/8af6473bc8c0f967e60353b49388808c5cacb66e)

crates/vm2/src/world_diff.rs
popzxc: Fixed in [01e089f](https://github.com/matter-labs/vm2/pull/95/commits/01e089f80b9891b975a7c84ef377b3b1b0917b37)

crates/vm2/src/world_diff.rs
popzxc: Fixed in [0e4ee2e](https://github.com/matter-labs/vm2/pull/95/commits/0e4ee2efe4e42fe180586dfd2efa2c8f267e374d)

crates/vm2/src/world_diff.rs
popzxc: Implemented in [27e291f](https://github.com/matter-labs/vm2/pull/95/commits/27e291f11523558ef0fa3789dbc74d654526a8ca)

popzxc: Implemented in [ac906df](https://github.com/matter-labs/vm2/pull/95/commits/ac906df8bb529d059cad68886de0f00d1681f791)

crates/vm2/src/instruction_handlers/decommit.rs
popzxc: This is a potential risk indeed, I'll look into that. Great catch!

crates/vm2/src/world_diff.rs
slowli: @popzxc Your summary of fast VM changes is completely correct; the execution divergence on OOG on decommitment was observed in the wild (on the Era mainnet IIRC).

crates/vm2/src/decommit.rs
```diff
@@ -2,11 +2,43 @@ use primitive_types::{H160, U256};
...
+};
+
+/// Ensures that a decommit hash has a materialized reusable page and returns it.
+///
+/// This page is pinned globally in [`WorldDiff`] and additionally kept alive in the
+/// bootloader frame (or current frame if no bootloader frame exists), matching
+/// decommit opcode teardown semantics.
+pub(crate) fn materialize_decommit_page<T: Tracer, W: World<T>>(
+    vm: &mut VirtualMachine<T, W>,
+    code_hash: U256,
```
shamatar, line 20: Generic comment - we are sensitive to passing large arguments by-value, so if there is some clippy option we may try to blindly pass all of them by-reference

crates/vm2/src/heap.rs
```diff
@@ -381,15 +381,7 @@ impl Heaps {
...
-        if page == HeapId::FIRST {
-            let prev_value = self[page].read_u256(start_address);
-            self.bootloader_heap_rollback_info
-                .push((start_address, prev_value));
-        } else if page == HeapId::FIRST_AUX {
-            let prev_value = self[page].read_u256(start_address);
-            self.bootloader_aux_rollback_info
-                .push((start_address, prev_value));
-        }
+        self.record_bootloader_word_rollback(page, start_address);
```
shamatar, line 384: What is a reason for bootloader's heap rollbacks? There are no memory rollbacks in proving, so it it a sequencing related thing? If so, should we ignore it (add some config flag and panic if rollback happens)?

popzxc: We have rollbacks in the code that is being ran for proving. Basically, how code works is that we try to run the code with bytecode compression, and if it fails, we re-run it without bytecode compression ([source](https://github.com/matter-labs/zksync-era/blob/main/core/lib/tee_verifier/src/lib.rs#L285)). 

In theory, it's pretty easy to solve -- just append data on whether compression should be used for a particular tx or not. In practice, I've measure the impact and it is not that big. Here's a sample of a big batch:
- Before all the heap preservation stuff: 40526990766 cycles.
- With all of that (including bootloader heap rollbacks): 40706137072 cycles.

So the difference is roughly 0.04%, IMHO it's negligible (while I agree that it's inefficient) -- it will let us to keep more code as-is and make the integration simpler.

Removing rollbacks and always executing with bytecode compression doesn't make the code significantly faster either (40667697931 vs 40706137072).

So idk, I think that introducing branching here & also changing more code to make it work is not justified, especially that we have ways to optimize code without sacrificing anything (e.g. by porting codebase to use ruint).

crates/vm2/src/decommit.rs
popzxc: Good point, but I probably won't do it in this PR and ship as a separate PR if required; diff here is huge already.

crates/vm2/src/state.rs
```diff
@@ -128,17 +150,21 @@ impl<T: Tracer, W: World<T>> State<T, W> {
...
+                self.heaps.deallocate(heap);
+            }
         }
         self.heaps.rollback(bootloader_heap_snapshot);
         self.registers = registers;
         self.register_pointer_flags = register_pointer_flags;
         self.flags = flags;
         self.transaction_number = transaction_number;
         self.context_u128 = context_u128;
+        self.next_base_page = next_base_page;
```
ninolipartiia, line 167: @popzxc, I think we should delete some slots from the `state.heaps.dynamic` that were added after the snapshot. Otherwise, there's a chance of panic later down the line in 
```rust
assert!(
    slot.is_none(),
    "heap page {} is already allocated",
     page.as_u32()
 );
```
See the issue [description](https://github.com/ninolipartiia/vm2/blob/popzxc-airbender-eravm-review/security-review/decommit-rollback-leak.md) and the [test](https://github.com/ninolipartiia/vm2/blob/popzxc-airbender-eravm-review/crates/vm2/src/tests/divergence_regressions.rs#L1290-L1546).

crates/vm2/src/instruction_handlers/precompiles.rs
ninolipartiia, line 85: @popzxc, should we use saturated/checked arithmetic here (cast_possible_wrap)?
It's not obvious to me whether it can cause any issues, nor if it's safe. I'd suggesting managing overflows and casts explicitly. Or alternatively please add a comment why is it safe

ninolipartiia: Similarly, in the _crates/vm2/src/precompiles/legacy.rs_, the `query.value = self.input.heap.read_u256(start_word * 32);` without the validation of `start_word` seems risky. Can you confirm if the multiplication here is safe? I suppose it relies on the precompiles using `execute_partial_query` correctly

popzxc: On cast_possible_wrap -- it corresponds to [this logic](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L372-L375) and later [this](https://github.com/matter-labs/zksync-protocol/blob/main/crates/zk_evm/src/opcodes/execution/log.rs#L453). I assume that assertion is missing in vm2 because ABI comes from kernel/system contracts, so it can be trusted.

On multiplication -- I guess it can overflow in theory, but AFAIU, in practice, this value is set by system contracts, so we _might_ say that it's expected that system contracts are correct. Again, as far as I understand, adding an assert _technically_ will be a divergence, since zk_evm seems to handle that case without overflow, but I'm just not sure if it's worth "proper" alignment. 

In general, it looks like vm2 cannot represent words beyond u32::MAX byte offset, not sure if we care about that. So we can either:
- add an assert here -- such a big value is certainly unexpected and probably something is wrong; but technically it's a divergence with zk_evm, or
- return 0 if multiplication overflows with `checked_mul` -- technically it's as close as we can get to zk_evm behavior without changing memory model.

I would probably prefer not to change the memory model over this, though, unless this fact brings us more potential and user-exploitable divergencies?

crates/vm2/src/instruction_handlers/heap_access.rs
```diff
@@ -190,6 +194,62 @@ fn load_pointer<T: Tracer, W: World<T>, const INCREMENT: bool>(
...
+        }
+
+        let address = pointer.low_u32();
+        let value = Register2::get(args, &mut vm.state);
+        vm.state
+            .heaps
+            .write_u256(STATIC_MEMORY_HEAP, address, value);
+
+        if INCREMENT {
+            Register1::set(args, &mut vm.state, pointer + 32);
```
ninolipartiia, line 248: @popzxc Shouldn't there be `Register2::set_fat_ptr` in case we need to preserve the pointer flag? That is, depending on the output from `get_with_pointer_flag`.

I think the same applies to `load`.

See the issue [description](https://github.com/ninolipartiia/vm2/blob/popzxc-airbender-eravm-review/security-review/01-static-memory-pointer-flag-divergence.md) and a minimal [test](https://github.com/ninolipartiia/vm2/blob/popzxc-airbender-eravm-review/tests/differential-regressions/src/tests/heap_read_increment_pointer_flag.rs).

ninolipartiia (approved):

0xVolosnikov (approved):
We can merge it and continue reviews/bugfixes in separate PRs

popzxc:
I will merge this PR and will work on remaining issues in follow up PRs.

PR #96, opened 2026-05-13
fix: Properly recycle heap pages on snapshot rollback

popzxc:
Context: https://github.com/matter-labs/vm2/pull/95#discussion_r3195221995

In snapshots, we now record the heap dynamic length, and on rollback, we properly recycle pages to get back to the same length.


ninolipartiia (approved):

0xVolosnikov (dismissed):
Theoretically we can have a pointer to a page that is already in `heaps_i_am_keeping_alive` and recycle it keeping the "dangling" pointer. But I don't think that this is achievable in practice

0xVolosnikov (approved):

PR #99, opened 2026-05-13
fix: Use empty heap for reading unallocated heaps

popzxc:
Report: https://github.com/ninolipartiia/vm2/blob/popzxc-airbender-eravm-review/security-review/04-failed-far-call-r1-poison-host-panic.md
Relevant code in zk_evm: https://github.com/matter-labs/zksync-protocol/blob/4a434874c9c7a306ed4bf807c55a86de29e47976/crates/zk_evm/src/reference_impls/memory.rs#L110-L115

On unallocated heap access for reads, return empty heap instead of panicking.

### DIFF
--- a/crates/vm2/src/heap.rs
+++ b/crates/vm2/src/heap.rs
@@ -398,26 +398,16 @@ impl Heaps {
     pub(crate) fn write_u256(&mut self, page: HeapId, start_address: u32, value: U256) {
         self.record_bootloader_word_rollback(page, start_address);
         let decoded = DecodedPage::decode(page)
-            .unwrap_or_else(|| panic!("heap page {} is not allocated", page.as_u32()));
-        let (heap, pagepool) = self
-            .try_decoded_page_mut(decoded)
-            .unwrap_or_else(|| panic!("heap page {} is not allocated", page.as_u32()));
+            .unwrap_or_else(|| panic!("heap page {} is not decodable", page.as_u32()));
+        let (heap, pagepool) = self.decoded_page_mut_for_write(decoded);
         heap.write_u256(start_address, value, pagepool);
     }
 
     pub(crate) fn write_bytes(&mut self, page: HeapId, start_address: u32, bytes: &[u8]) {
         self.record_bootloader_range_rollback(page, start_address, bytes.len());
         let decoded = DecodedPage::decode(page)
-            .unwrap_or_else(|| panic!("heap page {} is not allocated", page.as_u32()));
-        if let DecodedPage::Dynamic { .. } = decoded {
-            let slot = decoded_dynamic_slot_mut(&mut self.dynamic, decoded);
-            if slot.is_none() {
-                *slot = Some(Heap::default());
-            }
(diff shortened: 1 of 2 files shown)


ninolipartiia:
The fix looks good to me. 

The only thing I'd suggest is maybe consider avoiding other panics in the heap.rs, especially the similar instances of panics on unsuccessful page decode. I'll comment on a few instances shortly.

crates/vm2/src/heap.rs
ninolipartiia, line 407: The `write_to_memory` in the `zk_evm` ensures that the page exists instead of aborting.

https://github.com/matter-labs/zksync-protocol/blob/4a434874c9c7a306ed4bf807c55a86de29e47976/crates/zk_evm/src/reference_impls/memory.rs#L121-L125

ninolipartiia, line 420: Same as for the `write_u256`, the `zk_evm` ensures that the page exists instead of aborting.

https://github.com/matter-labs/zksync-protocol/blob/4a434874c9c7a306ed4bf807c55a86de29e47976/crates/zk_evm/src/reference_impls/memory.rs#L121-L125

ninolipartiia, line 387: I think the `zk_evm` does not have the page deallocate functionality, so maybe we should avoid panics in `deallocate` in `vm2` as well to avoid divergences. But let me know what you think.

popzxc: If we panic in deallocation, it means that there is a bug in our internal bookkeeping, so I think it's pretty much the right place to panic.

popzxc:
@ninolipartiia addressed panics that can theoretically happen; there were more cases where original VM doesn't panic but our does (memory-related), but these seem to be inaccessible under assumptions that system contracts work as expected, so IMO not worth fixing.

crates/vm2/src/heap.rs
ninolipartiia, line 367: @popzxc, wanted to flag this assert. I think it might make sense to remove it. This PR allows `write_u256` and `write_bytes` to allocate pages lazily (to match `zk_evm` behaviour), which could mean this assert no longer holds as a strict invariant. Alternatively, we could check that the slot is either None or an empty heap that was lazily allocated.
Either way, the panic doesn't seem exploitable, more of a consistency note than a blocker. Also, everything else looks good to me.

0xVolosnikov: iiuc currently unreachable, but will be flagged during reviews

0xVolosnikov: iiuc currently unreachable, but will be flagged during reviews

0xVolosnikov: I think we can keep it, but would be nice to reflect such details in comments to not waste the future reviewers time

0xVolosnikov (dismissed):
Generally lgtm, but would be nice to add comments about unreachable panics/issues

0xVolosnikov (approved):

PR #102, opened 2026-05-28
chore: v31 ooc patch

0xVolosnikov:


crates/vm2/src/instruction_handlers/ret.rs
```diff
@@ -213,11 +217,11 @@ impl<T: Tracer, W: World<T>> Instruction<T, W> {
...
 
     /// Creates a panic [`Ret`](opcodes::Ret) instruction with the provided params.
-    pub fn from_panic(label: Option<Immediate1>, arguments: Arguments) -> Self {
+    pub fn from_panic(src1: Register1, label: Option<Immediate1>, arguments: Arguments) -> Self {
         let to_label = label.is_some();
         Self {
             handler: monomorphize!(ret [T W Panic] match_boolean to_label),
-            arguments: arguments.write_source(&label),
+            arguments: arguments.write_source(&src1).write_source(&label),
         }
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Acknowledged. `zksync_vm2` is still on `0.x` (currently `0.5.0`), so a signature change here is within SemVer policy. The break is intentional rather than a backwards-compatible overload because the new parameter is load-bearing for the opcode (the panic ABI source) and a wrapper that hardcodes `Register1(Register::new(0))` would silently produce a different observable VM state than what the bytecode encodes. I will flag it in the next release CHANGELOG entry instead.

crates/vm2/src/tests/divergence_regressions.rs
```diff
@@ -1083,3 +1083,322 @@ fn rollback_should_restore_bootloader_heap_after_fresh_decommit() {
...
+
+    // The zero-ABI run should leave most of the forwarded gas intact.
+    assert!(
+        ergs_with_zero_abi > initial_gas - gas_to_pass,
+        "zero-ABI run drained more than expected: {ergs_with_zero_abi}"
+    );
+
+    let diff = ergs_with_zero_abi.saturating_sub(ergs_with_overflow_abi);
+    assert!(
+        diff > gas_to_pass - 100,
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Done in 3869197 — extracted `CALLEE_OVERHEAD_BOUND = 100` with a comment explaining what it covers (the callee runs one `add` reading from its code page at 6 ergs and a terminal opcode at 5 ergs, plus framework slack). Same constant is now used in the assertion margin.

crates/vm2/src/tests/panic.rs
```diff
@@ -17,7 +17,7 @@ proptest! {
                 Immediate2(0xFFFF),
                 Arguments::new(Predicate::Always, 25, ModeRequirements::none()),
             ),
-            Instruction::from_panic(Some(Immediate1(label)), Arguments::new(Predicate::Always, 5, ModeRequirements::none())),
+            Instruction::from_panic(Register1(Register::new(0)), Some(Immediate1(label)), Arguments::new(Predicate::Always, 5, ModeRequirements::none())),
```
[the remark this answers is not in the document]
0xVolosnikov: [AI] Done in 3869197. `rustfmt` did not reformat it on its own (the call sits inside a `vec![]` macro where rustfmt is conservative about rewriting), so the wrapping is manual but matches the style of the surrounding `from_near_call(...)` call.

crates/vm2/src/instruction_handlers/binop.rs
```diff
@@ -36,6 +36,7 @@ where
     boilerplate::<Op, _, _>(vm, world, tracer, |vm, args| {
         let a = In1::get(args, &mut vm.state);
         let b = Register2::get(args, &mut vm.state);
+        vm.state.clear_dst1(args);
```
ninolipartiia: I suspect this should be moved to after `Out::set(args, &mut vm.state, result)`. This seems to be an edge case when `dst0_reg_idx == dst1_reg_idx`: in that case `Out::set` ends up overwriting the dst1, but I suspect we want to unconditionally clear dst1.

@0xVolosnikov , what do you think?

0xVolosnikov: This is a good point, let me check EraVM behavior

0xVolosnikov: [AI] I looked at the EraVM reference (the pinned `zk_evm` commit `55f7a4c`) and the current placement actually matches it. Concretely:

`zk_evm/src/vm_state/cycle.rs:462` runs the pre-zero, then `apply` is dispatched. For `Add`, `zk_evm/src/opcodes/execution/add.rs` ends with `vm_state.perform_dst0_update(..., result, dst0_mem_location, self)`. `perform_dst0_update` (`zk_evm/src/vm_state/helpers.rs:407-424`) writes through `update_register_value(opcode.dst0_reg_idx, value)` and does not touch `dst1_reg_idx`.

So when `dst0_reg_idx == dst1_reg_idx`, zk_evm sequence is:
1. `update_register_value(dst1_reg_idx, empty())` → reg X = 0
2. `add_opcode_apply` → `perform_dst0_update` → `update_register_value(dst0_reg_idx, result)` → reg X = result

Final reg X = `result`, not zero.

I verified this against my placement with a regression test I just pushed (`add_with_dst0_equal_to_dst1_should_yield_sum`, commit `056709e`): `add r1, r0, r3` with `dst1_reg_idx` patched to `r3` (= dst0), `r1 = 42`, expected `r3 = 42`. Passes with the current placement. To double-check, I locally moved the clear to after `Out::set` (your suggestion) and re-ran — the test fails with `left: 0, right: 42`, confirming that placement would diverge from zk_evm in the `dst0 == dst1` alias case.

So I think we want to keep the clear before the writes: matches zk_evm semantics for both `dst0 == dst1` (dst0 wins) and `src* == dst1` (sources read original values).

0xVolosnikov: [AI] Follow-up: my earlier reply only covered the case I tested (register dst0), but you were partially right — the placement *was* wrong for **stack-mode dst0**, just for a different reason than `Out::set` overwriting dst1.

The AFL fuzzer caught it after the dep bump. Concrete trace: a binop with `dst0 = stack[r15 + 1028]` (push form) and `dst1_reg_idx = 15`. vm2 (left) and zk_evm (right) diverged by 1 in the resulting SP (`33288` vs `33287`). Root cause:

- For stack destinations, `Destination::set` calls `destination_stack_address(args, state)` (`crates/vm2/src/addressing_modes.rs:358`), which reads the register at `args.destination_registers.register1()` — i.e. r15 — to compute the address.
- With the clear positioned **before** `Out::set`, that read returned the already-cleared `0`. The original r15 value (off-by-1 from zero in the fuzz input) was lost.
- zk_evm captures `dst0_mem_location` into prestate **before** the cycle prologue (`update_register_value(dst1_reg_idx, empty())` at `crates/zk_evm/src/vm_state/cycle.rs:463`), so its address still reflects the original r15.

So the ordering really is dst-type-dependent:
- **Register dst** (`Out::IS_REGISTER == true`): clear *before* `Out::set`. If `dst0 == dst1`, the write overrides the cleared zero — matches zk_evm.
- **Stack dst** (`Out::IS_REGISTER == false`): `Out::set` *before* clear. The address is computed from the still-original register; then the clear zeros it — matches zk_evm.

Pushed in [`62f51fd`](https://github.com/matter-labs/vm2/commit/62f51fd):
- Added `const IS_REGISTER: bool` to the `Destination` trait (`crates/vm2/src/addressing_modes.rs`).
- Reordered the clear in `binop` (`crates/vm2/src/instruction_handlers/binop.rs`) and `ptr` (`crates/vm2/src/instruction_handlers/pointer.rs`) based on `Out::IS_REGISTER`.
- `nop` already uses `AdvanceStackPointer` for both source and destination — moved the clear after the `destination_stack_address` call (`crates/vm2/src/instruction_handlers/nop.rs:18-22`).
- New regression test `add_with_stack_push_dst0_aliasing_dst1_should_use_original_register_for_address` exercises exactly the fuzz case; reverting the binop reorder makes it fail with `left: 100, right: 4342` (i.e. `0 + offset` vs `sentinel + offset`).

The dst0_reg_idx == dst1_reg_idx test I added in `056709e` for the register-dst case (assert: `r3 = 42`) still passes — that path is unchanged.

PR #104, opened 2026-06-14
perf: cut verifier memory — gate storage_logs trace (Boojum-safe) + consolidate WorldDiff maps

0xVolosnikov:
## What
Cuts `WorldDiff` memory for re-execution verifiers (consumer: matter-labs/eravm-airbender-verifier#18) **without affecting Boojum**. Three independent changes:

1. **Gate the per-access `storage_logs` / `rollback_storage_logs` trace** behind `WorldDiff::set_record_storage_logs` (default **ON**). The trace is what `circuit_sequencer_api::sort_storage_access_queries` consumes for the in-circuit storage argument, so Boojum witness generation is unaffected. A re-execution verifier with no in-circuit storage argument (Airbender) opts out and derives the deduplicated set from the rollback-aware maps instead — saving ~270 MiB on large batches.

2. **Consolidate the `(address, key)` maps** so the 52-byte key is stored once instead of up to five times:
   - the three membership sets (`read` / `written` / committed-read-at-depth-zero) → one `slot_flags: RollbackableMap<_, u8>` of bit flags (`SLOT_READ` / `SLOT_WRITTEN` / `SLOT_COMMITTED_READ_Z0`);
   - `storage_changes` (`U256`) + `paid_changes` (`u32`) → one `storage_writes: RollbackableMap<_, StorageWriteEntry { value, paid }>`.

   `SLOT_COMMITTED_READ_Z0` materializes the dedup's `did_read_at_depth_zero` predicate incrementally; `committed_reads_at_depth_zero_iter` + `initial_storage_value` expose what the verifier needs. Two single-traversal primitives keep the merge cycle-free on the hot path: `RollbackableMap::add_flags` (flag OR-merge in one traversal instead of get+insert) and `RollbackableMap::update` (journaling read-modify-write so `write_storage` never reads the entry back — `prepaid` comes from `insert`'s returned old value).

3. **Capacity-reservation helpers** (`reserve_auxiliary_log_capacity`, `Heaps::reserve_dynamic_groups` / `VirtualMachine::reserve_dynamic_heap_capacity`, `RollbackableLog::reserve`) to suppress `Vec`-doubling transients in the guest.

Combined: the worst-case production batch (67912) drops from needing ~920–952 MiB to **fitting at ~720 MiB**.

## Cycle impact
Net **cycle-neutral** — verified end-to-end on the Airbender guest (transpiler, batch 67912):

| `WorldDiff` | cycles | vs baseline |
|---|---|---|
| pre-consolidation (separate maps) | 100,958,847,193 | — |
| this PR (consolidated + both single-traversal opts) | 100,952,045,388 | **−6.8M (−0.007%)** |

The naive merge initially cost +238M cycles (the coupled `value`+`paid` entry forced a per-write read-back, and the merged flags map did get+insert per flag); the two single-traversal primitives above remove both, leaving the consolidated `WorldDiff` marginally **faster** than baseline while saving ~200 MiB. Output matches native verification at every build.

## Boojum safety
Recording stays on by default, so `storage_log_queries()` is byte-for-byte unchanged for existing consumers (read/write log values, timestamps, rollback twins, and the external-rollback `truncate` are only gated behind the opt-out, never altered). In recording mode `read_storage_inner` also matches the pre-optimization base on **memory** — read-only slots are not cached, so Boojum gains no extra map entries. **57 lib tests pass**, including the storage-log trace tests (`storage_logs_include_reads_writes_and_rollbacks`, `storage_log_uses_vm_transaction_number`, the rollback ones), the `storage_changes_*` proptests, the merged-entry `paid`+rollback test, and the opt-out test (`skip_storage_logs_drops_trace_but_keeps_dedup_inputs`). clippy + rustfmt clean.

## ⚠️ Public API change
`WorldDiff::get_storage_state()` now returns `&BTreeMap<_, StorageWriteEntry>` (was `…, U256>`), and `StorageWriteEntry` is re-exported from the crate root. **Direct callers must project `.value`.** The downstream eravm `vm_fast` consumer (2 sites) must take its vm2-pin bump together with that `.value` projection in one commit.

## Notes
- Branched on `master` (includes #99 / #96).
- Folds in the map-consolidation work originally split as #105 / #106 (now merged into this branch — those PRs can be closed).
- Supersedes #100 (same memory change, but unconditional — it broke the storage-log trace; this version gates it so Boojum is preserved).
- `TODO`: pin to a tag before downstream merge.

### DIFF
--- a/crates/vm2/src/heap.rs
+++ b/crates/vm2/src/heap.rs
@@ -222,6 +222,12 @@ impl DynamicPageGroup {
         self.code.is_none() && self.heap.is_none() && self.aux.is_none()
     }
 
+    fn recycle(self, pagepool: &mut PagePool) {
+        for heap in [self.code, self.heap, self.aux].into_iter().flatten() {
+            heap.recycle(pagepool);
+        }
+    }
+
     fn slot(&self, kind: DynamicPageKind) -> Option<&Heap> {
         match kind {
             DynamicPageKind::Code => self.code.as_ref(),
@@ -380,6 +386,10 @@ impl Heaps {
         heap.recycle(&mut self.pagepool);
     }
 
+    pub(crate) fn dynamic_len(&self) -> usize {
+        self.dynamic.len()
+    }
+
     pub(crate) fn write_u256(&mut self, page: HeapId, start_address: u32, value: U256) {
         self.record_bootloader_word_rollback(page, start_address);
         let decoded = DecodedPage::decode(page)
@@ -422,6 +432,19 @@ impl Heaps {
         for (address, value) in self.bootloader_aux_rollback_info.drain(aux_snap..).rev() {
             self.bootloader_aux_heap
                 .write_u256(address, value, &mut self.pagepool);
+        }
+    }
+
+    pub(crate) fn truncate_dynamic_to(&mut self, len: usize) {
+        assert!(
+            len <= self.dynamic.len(),
+            "cannot restore dynamic heap length {} from current length {}",
+            len,
+            self.dynamic.len()
+        );
+
+        for group in self.dynamic.drain(len..) {
+            group.recycle(&mut self.pagepool);
         }
     }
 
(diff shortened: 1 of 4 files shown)


shamatar (dismissed):

shamatar:
If we agree that small changes are ok, then even larger saving should be achievable by merging all (address, key) -> something maps - it's 20 + 32 duplicated bytes per storage slot into 3 maps

PR #109, opened 2026-07-02
perf: cut verifier memory — gate storage_logs trace (Boojum-safe) + consolidate WorldDiff maps

0xVolosnikov:
## Summary

Three independent optimizations to reduce WorldDiff verifier memory:

1. Gate `storage_logs` recording (dual-mode) — conditional per-access trace
   recording, Boojum-memory-identical, opt-out saves ~270 MiB on large batches.
2. Consolidate WorldDiff `(address, key)` maps — merge duplicate keyed
   collections into unified entries with bit flags.
3. Capacity reservation helpers — pre-allocate to suppress mid-execution
   vector-doubling transients.

## Impact

- Worst-case production batch (size 67912): ~920–952 MiB → ~720 MiB (~200 MiB saved).
- Cycle-neutral (−0.007% measured).
- Public API: downstream consumers project `.value` from the new `StorageWriteEntry` type.

## Changes

Net diff vs `master`: 7 source files (`world_diff.rs`, `rollback.rs`, `heap.rs`,
`vm.rs`, `tracing.rs`, `lib.rs`, `single_instruction_test/heap.rs`).
`Cargo.lock` unchanged; no dependency changes.

### DIFF
--- a/crates/vm2/src/world_diff.rs
+++ b/crates/vm2/src/world_diff.rs
@@ -74,22 +74,16 @@ pub struct WorldDiff {
     // This is never rolled back. It is just a cache to avoid asking these from DB every time.
     storage_initial_values: BTreeMap<(H160, U256), StorageSlot>,
 
-    /// Selects between two mutually exclusive bookkeeping modes, configured once
-    /// before execution via [`Self::set_record_storage_logs`]; never rolled back.
+    /// Selects two mutually exclusive bookkeeping modes (see
+    /// [`Self::set_record_storage_logs`] for the rationale); set once before
+    /// execution, never rolled back.
     ///
-    /// Defaults to `false` (recording on): `read_storage_inner` / `write_storage`
-    /// append the per-access `storage_logs` / `rollback_storage_logs` trace and
-    /// otherwise behave exactly like the pre-optimization base (read-only slots
-    /// are *not* cached in `storage_initial_values`). This is what consumers that
-    /// build an in-circuit storage argument from `storage_log_queries()` need
-    /// (e.g. Boojum witness generation via `sort_storage_access_queries`).
-    ///
-    /// When `true`, the trace is dropped and instead the inputs a re-execution
-    /// verifier needs are recorded: the `SLOT_COMMITTED_READ_Z0` flag and the
-    /// read-only `storage_initial_values` cache. A verifier with no in-circuit
-    /// storage argument (e.g. Airbender) derives the deduplicated set from the
-    /// flag + `storage_writes`, dropping the trace's memory cost (~270 MiB on
-    /// large batches) without adding any per-read map entries to Boojum's path.
+    /// - `false` (default): append the `storage_logs` / `rollback_storage_logs`
+    ///   trace; otherwise behave like the pre-optimization base — read-only slots
+    ///   are *not* cached in `storage_initial_values`.
+    /// - `true`: drop the trace; instead record the `SLOT_COMMITTED_READ_Z0` flag
+    ///   and the read-only `storage_initial_values` cache, the inputs a
+    ///   re-execution verifier derives the deduplicated set from.
     skip_storage_logs: bool,
 }
 
@@ -131,12 +125,9 @@ impl WorldDiff {
     /// flag + `storage_writes` instead, can pass `false` to avoid the trace's
     /// memory cost (~270 MiB on large batches).
     ///
-    /// Must be called before any storage access; toggling mid-execution would
-    /// leave a partial trace (recording) or partial dedup state (opt-out).
-    /// Enforced by the assertion below rather than left to the caller.
-    ///
     /// # Panics
-    /// Panics if any storage slot has already been read or written.
+    /// Panics if any storage slot has already been read or written — toggling
+    /// mid-execution would leave a partial trace or partial dedup state.
     pub fn set_record_storage_logs(&mut self, record: bool) {
         assert!(
             self.storage_logs.is_empty()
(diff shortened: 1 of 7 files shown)


crates/vm2/src/world_diff.rs
```diff
@@ -123,30 +204,57 @@ impl WorldDiff {
...
+                // No pending write at read time: `did_read_at_depth_zero`.
+                self.slot_add_flag((contract, key), SLOT_COMMITTED_READ_Z0);
+            }
+            live_write.unwrap_or(initial_value)
+        } else {
+            // Recording mode: read like the pre-optimization base, without
+            // caching read-only slots (keeps Boojum's memory unchanged).
+            self.just_read_storage(world, contract, key)
+        };
+        if !self.skip_storage_logs {
```
shamatar: It's hard to read, can you move it to the "else" branch right above?

crates/vm2/src/world_diff.rs
```diff
@@ -45,20 +61,38 @@ pub struct WorldDiff {
...
-    written_storage_slots: RollbackableSet<(H160, U256)>,
+    /// Per-slot access flags merged from three former `(address, key)` sets
+    /// (read / written / committed-read-at-depth-zero) so the 52-byte key is
+    /// stored once instead of three times. External-rollback semantics
+    /// (whole-VM only), matching the original sets. See the `SLOT_*` consts.
+    ///
+    /// `SLOT_COMMITTED_READ_Z0` keeps the dedup's `did_read_at_depth_zero`
+    /// predicate: set only by `read_storage_inner`, only in opt-out mode, and
+    /// only when `storage_writes` has no pending write for the slot at read time.
+    slot_flags: RollbackableMap<(H160, U256), u8>,
```
shamatar, line 76: I do not think it can be rollbackable map - e.g. an event of reading slot at 0 rollback depth should not be rolled back - it's a global property property.

shamatar (dismissed):

crates/vm2/src/world_diff.rs
0xVolosnikov: Fixed

crates/vm2/src/world_diff.rs
0xVolosnikov: It is rolled back only on external rollbacks (dropping the tx fully)

crates/vm2/src/world_diff.rs
```diff
@@ -80,6 +118,53 @@ pub(crate) enum DecommitState {
...
+    /// witness generation via `sort_storage_access_queries`). A re-execution
+    /// verifier with no in-circuit storage argument (e.g. Airbender), which
+    /// derives the deduplicated storage set from the `SLOT_PROTECTIVE_READ`
+    /// flag + `storage_writes` instead, can pass `false` to avoid the trace's
+    /// memory cost (~270 MiB on large batches).
+    ///
+    /// # Panics
+    /// Panics if any storage slot has already been read or written — toggling
+    /// mid-execution would leave a partial trace or partial dedup state.
+    pub fn set_record_storage_logs(&mut self, record: bool) {
```
ninolipartiia, line 135: It operates on `mut WorldDiff`, so I suppose it should not be called in the middle of VM execution?
Should other fields be checked to be empty?

crates/vm2/src/world_diff.rs
```diff
@@ -255,6 +385,26 @@ impl WorldDiff {
...
+            .as_ref()
+            .iter()
+            .filter(|(_, f)| **f & SLOT_PROTECTIVE_READ != 0)
+            .map(|(k, _)| *k)
+    }
+
+    /// Returns the initial (pre-batch) value of a slot if it has been
+    /// touched by a read or write during execution. Used by per-slot summary
+    /// derivation in place of walking the `storage_logs` trace.
+    pub fn initial_storage_value(&self, contract: H160, key: U256) -> Option<crate::StorageSlot> {
```
ninolipartiia, line 408: nit: it'd be good to document that this function behavior depends on the `skip_storage_logs` mode

crates/vm2/src/world_diff.rs
```diff
@@ -123,30 +208,58 @@ impl WorldDiff {
...
-            written_value: value,
-            rw_flag: false,
-            rollback: false,
-            is_service: false,
-        });
+        let value = if self.skip_storage_logs {
+            // Opt-out mode: no trace is kept, so record what the deduplicated set
+            // is derived from instead — cache the initial value on first read
+            // (writes already cache it) and flag a depth-zero read.
+            let initial_value = self
```
ninolipartiia: nit: `initial_value` can be moved inside `else` branch L:233

crates/vm2/src/world_diff.rs
0xVolosnikov: Correct — it must be called before any storage access (before VM execution begins), never mid-execution; the `assert!` enforces that. On the other-fields question: the guard is already sufficient because every storage read and write calls `slot_add_flag(...)` (in `read_storage_inner` and both branches of `write_storage`), so `slot_flags.is_empty()` alone proves no slot has been touched — the other three checks are just belt-and-suspenders.

crates/vm2/src/world_diff.rs
0xVolosnikov: Done in d94719c — added a doc note that the set of slots returning `Some` depends on the recording mode.

crates/vm2/src/world_diff.rs
0xVolosnikov: Done in d94719c — moved it into the `else` branch. It is only used there, and caching in the live-write branch was redundant since `write_storage` always caches the initial value first.

ninolipartiia (approved):

PR #110, opened 2026-07-02
fix: clear dst1 to zero

0xVolosnikov:
## Change

Mirrors zk_evm's mechanism

## zk_evm pin

Bumped zk_evm/zkevm_opcode_defs/zk_evm_abstractions to ac58f328 (v0.153.12)

### DIFF
--- a/crates/vm2/src/state.rs
+++ b/crates/vm2/src/state.rs
@@ -179,6 +179,11 @@ impl<T: Tracer, W: World<T>> State<T, W> {
         self.transaction_number = transaction_number;
         self.context_u128 = context_u128;
         self.next_base_page = next_base_page;
+
+        // Transient per-instruction bookkeeping, not part of the snapshot. It is always reset at the
+        // start of the next instruction before being read, so this is a functional no-op; we clear
+        // it here to keep the invariant "never survives an instruction boundary" self-evident.
+        self.dst1_was_updated = false;
     }
 
     pub(crate) fn delete_history(&mut self) {
--- a/crates/vm2/src/state.rs
+++ b/crates/vm2/src/state.rs
@@ -28,6 +28,10 @@ pub(crate) struct State<T, W> {
     pub(crate) transaction_number: u16,
     pub(crate) context_u128: u128,
     pub(crate) next_base_page: u32,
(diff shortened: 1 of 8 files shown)


ninolipartiia:
@0xVolosnikov, the [issue](https://github.com/ninolipartiia/vm2/blob/popzxc-airbender-eravm-review/security-review/08-dst1-register-not-pre-zeroed.md) included a part about the gas-accounting divergence for the `ret.panic`. I think it's not addressed here, is it?

0xVolosnikov:
@ninolipartiia ah, somehow missed this one. Will implement in a separate PR

crates/vm2/src/state.rs
```diff
@@ -28,6 +28,10 @@ pub(crate) struct State<T, W> {
     pub(crate) transaction_number: u16,
     pub(crate) context_u128: u128,
     pub(crate) next_base_page: u32,
+    /// Whether the current instruction has written its second output (`dst1`) register.
+    /// Set by every `dst1` write; if still `false` after execution, `dst1` is cleared to zero,
+    /// matching `zk_evm`. Transient per-instruction bookkeeping; excluded from equality and snapshots.
+    pub(crate) dst1_was_updated: bool,
```
ninolipartiia, line 34: Just want to be sure, why should `dst1_was_updated` not be included in the `PartialEq` impl? 

Also, omitting it from `StateSnapshot` makes sense, but I think it'd be clearer if `dst1_was_updated` were reset to false on rollback, wdyt? 

Besides that, LGTM.

0xVolosnikov: [AI] Thanks!

On `PartialEq`: intentionally excluded. `State::eq` is the differential-test comparator against a `zk_evm`-derived state, and `dst1_was_updated` is vm2-internal per-instruction bookkeeping with no `zk_evm` counterpart — including it could only produce spurious inequalities. It is also reset to `false` at the top of every instruction (`common.rs`) before it is read again, so its value at any comparison point carries no meaning.

On resetting it on rollback: good call — done in 0e0d056. It is a functional no-op (the next instruction resets it before any read), but clearing it in `rollback` keeps the "never survives an instruction boundary" invariant self-evident.

ninolipartiia (approved):

PR #112, opened 2026-07-07
fix: charge return-ABI heap growth on far ret.panic

0xVolosnikov:
## Summary

Follow-up to #110. The [security review issue](https://github.com/ninolipartiia/vm2/blob/popzxc-airbender-eravm-review/security-review/08-dst1-register-not-pre-zeroed.md) is **cumulative** and documents two PR #217 divergences; #110 fixes the `dst1` one (Divergence 1), this PR fixes the second: the `ret.panic` return-ABI heap-growth charge (Divergence 2), as flagged in [#110's review](https://github.com/matter-labs/vm2/pull/110#issuecomment).

**Stacked on `vv/dst1-clear`** because the fix relies on the post-#217 `zk_evm` pin that #110 introduces — landing it against the pre-#217 pin would make vm2 newly diverge from its `single_instruction_test` differential reference. Retarget to `master` once #110 merges.

## The divergence

On a **far** return the proving circuit and post-#217 `zk_evm` parse the return-ABI fat pointer from the raw source register and charge implied heap growth **even on panic**. A crafted fresh-heap pointer (`UseHeap`, `start = 0xFFFFFFFF, length = 1`) is non-addressable → `u32::MAX` growth → the frame's ergs drain to zero.

vm2 short-circuited to `None` on `ReturnType::Panic` and never parsed the return ABI, so the callee kept its gas and returned it to the caller. That's observable state the stricter circuit cannot reproduce → unprovable batch / differential mismatch against post-#217 `zk_evm`. Same MEDIUM class as #110 (vm2 is *more permissive* — no theft risk).

## Fix

Resolve the return-ABI register for **cost** on the panic path, passing `already_failed = true` so `get_calldata` charges the growth penalty while discarding the (unused) returndata pointer. This mirrors the circuit's "erase returndata, keep `upper_bound`" split and how vm2's own `ret.ok`/`ret.revert` already charge via `get_calldata`.

I verified against the pinned post-#217 `zk_evm` source (`ret.rs`): on `ret.panic` it empties the pointer (→ no growth for a *valid* pointer) but keeps the pre-erasure validation exceptions, so an overflowing pointer still triggers the `u32::MAX` penalty. `already_failed = true` reproduces exactly this (valid → `None` before `grow(bound)`; overflow → `grow(u32::MAX)`).

Plumbing: `from_panic` and the decoder previously dropped the source register (the value was never read); they now pass it through.

## Testing

- New regression test `far_ret_panic_charges_heap_growth_for_overflowing_pointer` in `divergence_regressions.rs`: a far callee does `ret.panic r3` with an overflowing fresh-heap pointer; asserts the callee's gas was drained (caller receives ~nothing back). Written test-first — confirmed it **fails** before the fix (caller keeps 69795 gas) and **passes** after.
- Full suite: 57 tests pass, clippy clean, `cargo fmt` clean, builds under `--features single_instruction_test`.

### DIFF
--- a/crates/vm2/src/instruction_handlers/ret.rs
+++ b/crates/vm2/src/instruction_handlers/ret.rs
@@ -41,17 +41,16 @@ fn naked_ret<T: Tracer, W: World<T>, RT: TypeLevelReturnType, const TO_LABEL: bo
 
         (snapshot, near_call_leftover_gas)
     } else {
+        let (raw_abi, is_pointer) = Register1::get_with_pointer_flag(args, &mut vm.state);
         let return_value_or_panic = if return_type == ReturnType::Panic {
             // A panic forwards no returndata, but the return-ABI pointer must still be resolved for
             // its heap-growth cost: a fresh-heap pointer whose `start + length` overflows `u32`
             // grows the heap to `u32::MAX`, draining the frame's gas. Passing `already_failed = true`
             // charges exactly that penalty while discarding the (unused) returndata pointer. This
             // mirrors the proving circuit and post-#217 zk_evm; see `get_calldata`.
-            let (raw_abi, is_pointer) = Register1::get_with_pointer_flag(args, &mut vm.state);
             get_calldata(raw_abi, is_pointer, vm, true);
             None
         } else {
-            let (raw_abi, is_pointer) = Register1::get_with_pointer_flag(args, &mut vm.state);
             let result = get_calldata(raw_abi, is_pointer, vm, false).filter(|pointer| {
                 if vm.state.current_frame.is_kernel {
                     true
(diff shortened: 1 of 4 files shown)


crates/vm2/src/instruction_handlers/ret.rs
```diff
@@ -42,6 +42,13 @@ fn naked_ret<T: Tracer, W: World<T>, RT: TypeLevelReturnType, const TO_LABEL: bo
         (snapshot, near_call_leftover_gas)
     } else {
         let return_value_or_panic = if return_type == ReturnType::Panic {
+            // A panic forwards no returndata, but the return-ABI pointer must still be resolved for
+            // its heap-growth cost: a fresh-heap pointer whose `start + length` overflows `u32`
+            // grows the heap to `u32::MAX`, draining the frame's gas. Passing `already_failed = true`
+            // charges exactly that penalty while discarding the (unused) returndata pointer. This
+            // mirrors the proving circuit and post-#217 zk_evm; see `get_calldata`.
+            let (raw_abi, is_pointer) = Register1::get_with_pointer_flag(args, &mut vm.state);
```
ninolipartiia: nit: `get_with_pointer_flag` can be moved before the `if` since it's computed in both branches

crates/vm2/src/instruction_handlers/ret.rs
ninolipartiia: After this PR's change, this comment is wrong

crates/vm2/src/instruction_handlers/ret.rs
0xVolosnikov: [AI]: Done — hoisted the `get_with_pointer_flag` call above the `if` in eb0a281.

crates/vm2/src/instruction_handlers/ret.rs
0xVolosnikov: [AI]: Fixed in eb0a281. Updated the comment to reflect that the args' source register *is* now read on the panic path for the return-ABI heap-growth charge; a spontaneous free panic passes empty args (register r0), so it reads zero and charges no growth.

ninolipartiia (dismissed):

ninolipartiia (approved):

PR #116, opened 2026-07-16
perf(heap): lazily-allocated 256B chunks + returndata-window compaction

0xVolosnikov:
## Problem
`HeapPage` was a dense `Box<[u8; 4096]>`, so touching one byte materialized a full 4 KiB page, and retained (returndata) heaps were kept whole. A "write one byte near a page boundary, keep the pointer" pattern allocates far more than it writes, and repeated across calls reaches multiple GB.

## Change
- **Sparse chunked storage.** Each heap is a flat `Vec<Option<Box<[u8; 256]>>>` indexed by `address >> 8` (`HEAP_CHUNK_SIZE`); an absent chunk reads as zero. A boundary write costs one 256 B chunk plus an 8 B index slot instead of a 4 KiB page. `ChunkPool` zeroes chunks on recycle so pooled memory retains nothing.
- **Returndata-window compaction.** On `pop_frame`, a bounded fat pointer can only ever address `[start, start + length)` of the heap it keeps alive (EraVM pointers narrow, never widen), so free every chunk outside that window back to the pool. Decommit-pinned code pages (shared) are left intact. Retained memory then tracks the bytes actually written and returned, not whole grown heaps.

## Impact
- Guest cycles **−1.4% total / −2.1% `vm_execution`** vs a page-based sparse heap on the 49-batch v31 corpus (the flat index removes an indirection on every heap access; the per-return compaction is negligible) — roughly **parity with the original dense heap** while keeping the memory bound.

### Adversarial scope (honest)
This does **not** lower the heap DoS worst case (~0.9 GB, near the 952 MiB guest cap). The optimal attack fills the free per-frame page (`NEW_FRAME_MEMORY_STIPEND = 4096 B`) with real writes and returns all of it, so there is nothing to compact — retained memory is ≈ bytes actually written, about the best a representation can do. The binding amplifier (free stipend × far-call count) is gas/protocol, not the host representation; closing the residual needs a protocol memory bound or a raised guest cap, not a heap-representation change.

## Tests
Differential fuzz vs a dense oracle (reads/writes/ranges incl. cross-chunk and cross-page, `from_bytes`, recycle-zeroing, absent-vs-present-zero equality); `retain_window` and `Heaps::compact_to_window` unit tests; `pop_frame_compacts_kept_heap_to_returndata_window` integration test (written test-first); adversarial worst-case characterization tests (`free_page_fill_is_not_reduced_by_compaction`, `compaction_reclaims_scratch_beyond_returned_window`). Full `zksync_vm2` suite, clippy `--all-targets`, and the `single_instruction_test` build all pass.


ninolipartiia (approved):

deniallugo-ml (approved):
Blind approve

PR #124, opened 2026-08-06
perf(stack): clear sub-chunks in place on reuse instead of freeing them

ninolipartiia:
## The problem

PR #115 makes the stack sparse: 4096 lazily-allocated 16-slot sub-chunks instead of one
dense 2 MiB `[U256; 65536]`. The sparseness is worth having; the recycling path is not.

`Stack::zero()` runs from `StackPool::get()` on every frame push that reuses a pooled
stack, and it *drops* every materialized sub-chunk — the callee's first write to each then
allocates it again. One far call can therefore cost up to 4096 frees plus 4096
`alloc_zeroed` calls, where the dense implementation did one allocation per stack and a
plain memset of dirty areas on reuse, with no per-chunk allocator traffic at all.

Per sub-chunk the trade is backwards: the free + `alloc_zeroed` round-trip costs ~500–1100
cycles against the ~64–128 of the 512 B memset it avoids — 5–10× the work it saves.

## Why this is worth fixing rather than tolerating

The allocator traffic is host-side work EraVM does not charge gas for, so it breaks the
proportionality gas pricing assumes. A contract that scatter-writes one slot per sub-chunk
and returns costs ~24,800 ergs per iteration (183 far call + 4096 × 6 ergs + 5 ret); in a
flat loop at the 80M-erg `MAX_TX_ERGS_LIMIT` that is ~3,200 iterations, i.e. **~26M
allocator operations in one transaction**. At 100–1000 cycles each that is
2.6×10⁹–2.6×10¹⁰ cycles, 3.8–38% of the 2³⁶ guest budget, against ≈1.7×10⁹ for the dense
stack's memset on the same pattern — an amplification of **2.5–16×** the attacker gets for
free, since they pay only for write opcodes they'd pay anyway.

Budget limits are a poor substitute: restoring the old margin that way means cutting the
budget by roughly that amplification factor, and assuming its pessimistic end because the
per-operation cost isn't knowable in advance — a permanent tax on honest heavy
transactions. Lowering the *per-tx* limit doesn't even help for batches, where the same
work-per-erg ratio simply spreads across more transactions.

Fragmentation is the part that cannot be calibrated against at all. The guest allocator is
a coalescing free-list (`talc`), and ~26M interleaved 512 B alloc/free cycles can degrade
every later allocation — heap chunks, program decodes, witness structures — with no
analytic bound.

## The change

All in `crates/vm2/src/stack.rs`:

1. **`zero()` clears materialized sub-chunks in place** (`chunk.fill(U256::zero())`) rather
   than dropping them, so after a stack's first materialization the reuse path performs no
   allocator work. A `debug_assert!` pins the invariant this relies on: nothing nonzero
   survives `zero()`.
2. **`rollback()` reuses materialized chunks** and skips materializing any whose snapshot
   slice is all zeros. It previously allocated a fresh chunk for every sub-chunk of every
   dirty area unconditionally, so one revert pinned them for the VM's lifetime.
3. **Allocation failure aborts cleanly**: `zeroed_chunk()` and `Stack::new()` call
   `handle_alloc_error` instead of passing a null `alloc_zeroed` result to `Box::from_raw`,
   which is UB. This matters in the guest, where the heap is fixed and allocation failure
   is reachable — a clean abort is a liveness failure, whereas UB can mean continuing with
   a corrupt stack.
4. **Regression tests** for the above, including the rollback that *narrows* `dirty_areas`
   — the one transition that can strand a materialized chunk in a non-dirty area.

**Why it is behaviour-preserving.** An absent sub-chunk and an all-zero one are already
indistinguishable: `get()` returns zero for `None`, and `PartialEq` treats
`(Some(all-zero), None)` as equal. `dirty_areas` evolution is untouched, so snapshot,
rollback and equality semantics are unchanged; the invariant weakens only from *allocated ⊆
dirty* to *nonzero ⊆ dirty*, which `set()` (marks its area in the same call) and
`rollback()` (restores `dirty_areas` alongside the data) both maintain. Empirically,
delegation counts and VM feature counters are bit-identical to #115 on all 49 corpus
batches and on the stress fixture.

## Measurements

A/B in the Airbender verifier: both trees exported, guest rebuilt per arm,
`CARGO_INCREMENTAL=0`, with a distinct guest `app.bin` sha256 per arm as the guard that the
`[patch]` actually reached the binary. A115 = #115 tip, B115 = #115 + this change.

**Allocator traffic on the reuse path** (counting `GlobalAlloc`, reuse window only). Counts
are exact and deterministic; timings are single-shot with ~30–40% run-to-run spread, so
read them as magnitudes:

| workload | allocs | frees | ms |
|---|---|---|---|
| scatter-all (200 frames × 4096 sub-chunks) | 819,200 → **4,096** | 815,104 → **0** | 503 → **31** |
| partial-scatter (2000 × 256) | 512,000 → **256** | 511,744 → **0** | 92 → **24** |
| shallow (20,000 × 64 contiguous) | 80,000 → **4** | 79,996 → **0** | 16 → **8** |

**Guest cycles, 49-batch organic corpus:**

| metric | value |
|---|---|
| total effective cycles | 343,744,358,330 → 343,497,381,177 (**−0.072%**) |
| per-batch delta | min −0.109% / median −0.077% / max −0.013% — **49 improved, 0 regressed** |
| phase attribution | `vm_execution` **−0.117%**, carrying 99.9% of the delta; other phases below jitter |
| worst-batch ceiling use | 13.76% → 13.75% of 2³⁶ |
| delegations / features | **bit-identical, 49/49** |

`vm_execution` is the only phase that runs `StackPool::get → Stack::zero`, so the win
landing there is causally coherent rather than incidental. Against the pre-sparse baseline
(v0.6.0, same method) #115 alone measures **−1.887%** and with this change **−1.957%** —
the sparse stack is a cycle win in its own right, since dense frames pay to zero 2 MiB on
every reuse.

One counter-result, reported for completeness: on a merkle-dominated synthetic fixture
(900065) the total moves **+0.0076%** with `vm_execution` +0.0796% — below the ~0.03%
same-source-different-build jitter floor, and that fixture spends 30.7 G of its 39.3 G raw
cycles in `merkle_verification`, making it a weak probe for a stack-path change. It does
mean this is a win on organic batches, not a universal one.

**Native heap peaks, 52 batches:**

| phase peak | max Δ | batches worse |
|---|---|---|
| `load_decode`, `end_setup` | +0.0 MiB | 0/52 |
| `end_vm`, `final` | **+0.1 MiB** | 13/52 |

Worst absolute increase is +0.1 MiB on peaks of 82.6–222.2 MiB (+0.07%), the worst-batch
peak is unchanged at 222.2 MiB, and 39/52 are bit-identical. No batch improves, so on real
workloads memory is neutral rather than better. The increase appears only in `end_vm` — the
phase that owns `StackPool` — which is where it belongs if the measurement is sensitive to
this change at all.

## The trade-off

A pooled stack's memory becomes its **high-water mark**, the union of every sub-chunk it
has ever touched, instead of tracking the live frame. Two probes written to expose this
show 8 KiB → 2 MiB on a rotating write window, and +93% peak (1.34 → 2.58 MB) at call
depth 32 with moving windows. Three things bound what that can cost:

- **It does not raise #115's worst case.** A fully materialized sparse stack is 2,121,736 B
  against the dense stack's 2,105,352 B — the difference is the chunk index, 16 KiB on
  rv32, +0.8%. One frame touching every sub-chunk already materializes 2 MiB in both arms,
  and residency costs 6 ergs per write either way, so the erg price of pinning a given
  number of sub-chunks is unchanged. What's removed is only the need to scatter within a
  single frame.
- **Real workloads don't reach it.** Organic contracts write low, contiguous slots, so a
  stack's historical union is essentially its live footprint — hence the ≤ +0.1 MiB above.
  This is an adversarial-shape property, not a workload cost.
- **`rollback()` moves the other way**, retaining strictly less than #115 does. The two
  probes don't exercise rollback, so they show this change's worst case, not its average.

Separately: `zero()` now memsets every materialized chunk in the dirty areas, so reuse cost
tracks the high-water mark rather than the frame's own writes — bounded above by what the
dense stack paid for the same dirty mask. Making it exact would need per-sub-chunk dirty
tracking on the hot write path: more invasive, and deliberately not part of this PR. If the
retention itself proves to matter, the narrower option is trimming on `recycle` rather than
`get`, so a stack parked in the pool doesn't hold its historical union.


0xVolosnikov:
worst-batch ceiling use | 13.76% → 13.75% of 2³⁶
-- | --


👀

Does it have the specific attack vector in test cases? It should appear in the benchmarks if this fix is actually valid/relevant

crates/vm2/src/stack.rs
```diff
@@ -71,29 +75,42 @@ impl Stack {
...
+        // trigger on every far call. Clearing only dirty areas suffices: a
+        // sub-chunk becomes nonzero only via `set` (which dirties its area) or
+        // `rollback` (which restores `dirty_areas` alongside the data), so
+        // chunks in non-dirty areas are already all-zero.
         for i in 0..NUMBER_OF_DIRTY_AREAS {
             if self.dirty_areas & (1 << i) != 0 {
                 for sc in (i * SUBCHUNKS_PER_AREA)..((i + 1) * SUBCHUNKS_PER_AREA) {
-                    self.slots[sc] = None;
+                    if let Some(chunk) = self.slots[sc].as_mut() {
+                        chunk.fill(U256::zero());
```
0xVolosnikov, line 117: [AI] [P2] This amplification is real but bounded by the protocol-enforced per-transaction erg limit N, so the right action is to characterize the bound rather than treat it as an unbounded DoS. I reproduced the structural case on 88d284f: after one call materializes all 4,096 chunks, a reused frame that writes one slot in each of the 64 coarse areas makes zero() fill all 4,096 retained chunks; #115 visits only 64. The one-time priming cost is P = 183 + 4,096*6 + 5 = 24,764 ergs and each steady reuse costs E = 183 + 64*6 + 5 = 572 ergs. Therefore k(N) = max(0, floor((N - P) / E)) full 2 MiB clears, and using the PR estimate of 64–128 cycles per chunk gives C(N) in [262,144*k(N), 524,288*k(N)]. For N = 15M, k = 26,180, about 54.9 GB is cleared cumulatively, and the estimated clearing cost is 6.86–13.73B cycles, or about 10–20% of 2^36. At the pessimistic 128-cycle bound, clearing alone reaches 2^36 only near N = 74,997,948. Thus this does not demonstrate a one-transaction DoS at a hard 15M ceiling. Please add the exact prime-then-one-write-per-area Airbender fixture to validate the per-chunk estimate and reserve enough of the guest budget for all other phases; if that confirms margin, I consider this non-blocking.

crates/vm2/src/stack.rs
```diff
@@ -141,15 +158,21 @@ impl Stack {
...
+        // zeros — absent already reads as zero. Non-dirty areas are all-zero
+        // after the `zero()` above.
         for i in 0..NUMBER_OF_DIRTY_AREAS {
             if dirty_areas & (1 << i) != 0 {
                 for sc in (i * SUBCHUNKS_PER_AREA)..((i + 1) * SUBCHUNKS_PER_AREA) {
-                    let mut chunk = zeroed_chunk();
-                    chunk.copy_from_slice(&slots[sc * SUBCHUNK_SLOTS..(sc + 1) * SUBCHUNK_SLOTS]);
-                    self.slots[sc] = Some(chunk);
+                    let src = &slots[sc * SUBCHUNK_SLOTS..(sc + 1) * SUBCHUNK_SLOTS];
+                    if self.slots[sc].is_none() && src.iter().all(U256::is_zero) {
```
0xVolosnikov, line 194: [AI] [P3] The counterexample remains valid, but under the one-transaction threat model this is a documentation issue rather than a demonstrated DoS. Snapshotting only area 0, materializing all 4,096 chunks afterward, and rolling back leaves all 4,096 chunks allocated here; #115 drops the divergent allocations and restores 64 chunks. Therefore the PR body should not say rollback retains strictly less in all cases. VM-visible state is correct, direct stack rollback is bounded by one stack, and this change does not raise the absolute fully materialized one-transaction stack ceiling relative to #115. Please rewrite the claim as workload-dependent and retain the counterexample in the memory trade-off discussion; trimming is not required by this finding.

0xVolosnikov (dismissed):
[AI] I revalidated these findings against current head 88d284f and comparatively against PR #115 using focused executable regression tests. Both findings reproduced. I am requesting changes for the coarse-dirty clearing amplification; I did not find a VM-state semantic correctness issue.

0xVolosnikov (commented):
[AI] Revised assessment after clarifying that the security invariant is safety against one L1-enforced priority transaction under a hard maximum N. The coarse-dirty amplification is bounded by k(N) = max(0, floor((N - 24,764) / 572)) full clears; at N = 15M, the PR estimates put clearing at about 10–20% of 2^36, so there is no demonstrated one-transaction DoS. I have withdrawn the blocking review. The remaining non-blocking asks are an exact Airbender stress fixture to validate the bound and a correction to the rollback retention wording.

PR #125, opened 2026-08-06
fix(heap): compact a returndata window only on the dying frame's own page

ninolipartiia:
## Problem

#116 added returndata-window compaction to `pop_frame`: when a returning frame forwards a fat pointer,
every 256-byte chunk of the kept heap page outside `[start, start + length)` is freed, and freed chunks
read back as zero. It did that for whatever page the returned pointer named, without checking that the
page belonged to the frame being popped.

`naked_ret` deliberately lets a kernel-mode frame forward a pointer to any page, mirroring zk_evm. So a
kernel callee that ret-forwards its `calldata_heap` names an older frame's heap that is still live.
That frame reads its own memory through `HeapRead`, which is bounded by `heap_size` and by no fat
pointer, so it resumed reading zeros where zk_evm returns the real bytes. zk_evm never frees a memory
page at all, so this is a consensus divergence, and it raises no gas, bounds or panic signal. Forwarding
the calldata pointer unmodified is enough; no crafted narrow window is needed.

## Fix

Gate the compaction on the page being the dying frame's own `heap` or `aux_heap`. `MakeNewPointer`
returns satisfy that by construction. The decommit-pin check stays, because `CodeOracle` (0x8012)
ret-forwards a `Decommit` pointer into its own heap: that page is owned but read in full by other
frames, so the pin is what protects it.

`heaps_i_am_keeping_alive` is excluded on purpose. It is filled after the frame swap with whatever page
the *child* returned, so it can hold a live ancestor's heap, and covering it would reintroduce the bug
with a three-frame chain.

The cost is returndata bubbling: a page forwarded up several frames now stays at the producer's window
instead of being re-narrowed at each hop. That is gain we forgo, never worse than pre-#116. Worst-case
retained memory is unchanged at roughly 0.9 GB, because it is set by the free per-frame stipend times
the far-call count, which is a protocol property rather than a representation one.

## What this does not fix

Ownership is necessary for the compaction to be sound, but it is not sufficient.

`PrecompileCall`'s `memory_page_to_read` and the `read_heap_byte`/`read_heap_u256` tracer API both take
a raw page id with no pointer, so both see a compacted page as zeros, including a page the dying frame
owns and this gate still compacts. A kernel frame can put a dead frame's page id in a precompile ABI
and hash bytes outside the window, producing a digest zk_evm would not. I confirmed that with a
contract-driven test which reproduces identically on this PR's base. **That test is not part of this
PR**, so the behaviour is not demonstrated by the diff.

Nothing in vm2 prevents it; what keeps it out of production is era-contracts. Of the eight per-contract
`unsafePackPrecompileParams` copies, seven cannot express a page id at all, and `Keccak256.yul` fills
it from its own live calldata pointer, which forwarding can only narrow. Any kernel-address contract
that packs a foreign page id reopens it.

The keep-alive *deallocation* sinks in `pop_frame` and `reclaim_bootloader_returndata_heaps` share this
trigger and predate the compaction. They stay open, and their symptoms are worse: they can free a live
frame's whole page, or panic on a duplicate `deallocate` when the smuggled page is the resuming frame's
own heap.

## Tests

Nine cases in `divergence_regressions.rs`: the two chains that must not compact (immediate caller and
live grandparent), the non-kernel panic counterpart, downward forwarding that must still be allowed,
the owned `heap` and `aux_heap` cases that must still compact, the same through a real `ret` so the
wiring is covered and not only the sink, and the decommit-pin exception.

Two things reviewers should know:

- `pop_frame_compacts_kept_heap_to_returndata_window` is renamed to
  `pop_frame_does_not_compact_a_kept_heap_the_dying_frame_does_not_own` and its assertions are
  inverted, because under the new predicate a keep-alive page must survive. Its original positive
  property moved to `pop_frame_compacts_dying_frames_own_heaps_to_returndata_window`.
- `keepalive_deallocation_frees_live_grandparent_heap` asserts the current, defective keep-alive
  behaviour on purpose, so it will fail when that separate defect is fixed. Rewrite it then rather than
  relaxing it; a silent regression would be worse.


0xVolosnikov:
Nice catch!

0xVolosnikov (approved):

